PR #80, opened 2017-08-09
More contrast to .kiwi-messagelist-time

mestaritonttu:
It was too light. See also: http://contrastrebellion.com/


prawnsalad:
I've lightened your colour just a little as it doesn't want to distract from the main messages, but it's now darker than it was. Thanks!

PR #104, opened 2017-09-12
State browser channel popup fix

iwalkalone69:
When statebrowser sidebar is not on the top of browser's window, popups are not rendering in the correct position.

This fix tries to see if kiwi-statebrowser has some CSS top parameter and, when rendering kiwi-statebrowser-channel-popup, uses it to render popup in the correct position.

### DIFF
--- a/src/components/StateBrowser.vue
+++ b/src/components/StateBrowser.vue
@@ -106,9 +106,14 @@ export default {
                 this.popup_networkid = null;
                 this.popup_top = 0;
             } else {
+                let h = 0;
+                let sb = document.getElementsByClassName('kiwi-statebrowser')[0];
+                if (sb.style && sb.style.top) {
+                    h = parseInt(sb.style.top, 10);
+                }
                 this.popup_buffername = buffer.name;
                 this.popup_networkid = buffer.networkid;
-                this.popup_top = domY;
+                this.popup_top = domY - h;
             }
         },
         closeBuffer: function closeBuffer() {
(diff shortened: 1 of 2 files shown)


src/components/StateBrowser.vue
```diff
@@ -106,9 +106,14 @@ export default {
                 this.popup_networkid = null;
                 this.popup_top = 0;
             } else {
+                let h = 0;
+                let sb = document.getElementsByClassName('kiwi-statebrowser')[0];
```
prawnsalad: This already has a reference to the element by `this.$el`

src/components/StateBrowser.vue
```diff
@@ -106,9 +106,14 @@ export default {
...
+                if (sb.style && sb.style.top) {
```
prawnsalad: What if the `top` style is not on this element but its parent instead? Better to use `this.$el.getBoundingClientRect()` to get the `top` property instead (+ it will always be a number).

prawnsalad: Also, please add a comment to explain why it's applying this logic to the DOM. ie. `Some custom themes may push the statebrowser down, so apply any top to the popup also`.

prawnsalad:
On top of the comments left on the edits, two other notes:
* Don't use single letter var names. They don't mean anything to anybody reading the code.
* That unrelated .gitignore change needs moving out of this PR

iwalkalone69:
Changes done as suggested:
- Rolled back modifications to .gitignore.
- Using this.$el.getBoundingClientRect() to get the top position of .kiwi-statebrowser

Some custom themes may push the statebrowser down, so apply any top to the popup also.

PR #115, opened 2017-09-27
Language autodetection

iwalkalone69:
This tries to load language based on browser settings. If it fails or navigator.language does not exist, it tries to load the language specified in settings.


prawnsalad:
Not sure what happened here but this code was merged in already. Thanks!

PR #125, opened 2017-11-03
Typo correction in app.dev.po and app.en-US.po (space seperated)

OndrejBakan:


prawnsalad:
Good catch. We actually use a translation manager for these files at http://translations.kiwiirc.com so I've updated it there.

Thanks!

Issue #89, opened 2017-08-11
Move focus to password field after ticking I have a password


mestaritonttu:
Smoother experience this way.

PR #136, opened 2017-11-21
Github Issue #89. 

Zappatta:
When Checking "I have a password", automatically set focus to the password input.

Created a new component called "optional-password", so code can be reused for both Welcome and CustomServer startup screens.


prawnsalad:
Thanks for this, I'm a bit late coming back here but we now have an easier focus directive that we can now use on elements which si a bit easier to use. https://github.com/kiwiirc/kiwiirc/commit/2d6c72efb0ea614476f0b4db44973a3b473190f1

Issue #135, opened 2017-11-21
Option to disable nick colouring in messages


prawnsalad:

Zappatta:
please see this pull request #137

PR #137, opened 2017-11-22
Issue #135

Zappatta:
Pretty self explanatory. Added flag in options, and checking flag before colouring nickname.

### DIFF
--- a/src/components/MessageList.vue
+++ b/src/components/MessageList.vue
@@ -234,7 +234,10 @@ export default {
             return highlightFound;
         },
         nickStyle: function nickColour(nick) {
-            return 'color:' + TextFormatting.createNickColour(nick) + ';';
+            if (state.setting('buffers.colour_nicknames_in_messages')) {
+                return 'color:' + TextFormatting.createNickColour(nick) + ';';
+            }
+            return '';
         },
         onListClick: function onListClick(event) {
             this.toggleMessageInfo();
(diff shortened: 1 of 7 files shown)


src/components/MessageList.vue
```diff
@@ -234,7 +234,10 @@ export default {
             return highlightFound;
         },
         nickStyle: function nickColour(nick) {
-            return 'color:' + TextFormatting.createNickColour(nick) + ';';
+            if (state.setting('buffers.colour_nicknames_in_messages')) {
```
prawnsalad: This should use `this.bufferSetting('colour_nicknames_in_messages')` so that per-buffer settings are possible.

src/res/locales/app.he-IL.po
```diff
@@ -166,6 +166,11 @@ msgstr "הודעות בגלילה לאחורה"
 msgid "settings_formatting"
 msgstr "עיצוב נוסף"
 
+#: Nickname Colouring
+msg "settings_nick_colouring"
+msgstr: "צבע את כינויי המשתמשים בהודעות"
+
+
```
prawnsalad: Only add the translations to `app.dev.po`. The rest are generated by translations.kiwiirc.com by using `app.dev.po` as a template

Zappatta:
@prawnsalad requests fulfilled. I will also modify the languages in the web interface. Please let me know if anything else is needed.

prawnsalad:
Nice, thanks!

Issue #110, opened 2017-09-22
Whois channels with modes not clickable


ManiacTwister:
In the /whois output all channels where the whoisd user has an "special" channel mode (like op, voice, ,,) are not clickable.
![kiwiibug](https://user-images.githubusercontent.com/893193/30733138-f96a26ae-9f75-11e7-91f0-54a34f8f5ba4.png)

prawnsalad:
This is now fixed in master. Thanks!

PR #138, opened 2017-11-24
Issue #110. Channel with user modes (received in whois text) were not "linkified"

Zappatta:


prawnsalad:
Few notes here:
* The channel match group already exists as an argument
* `\007` is purposely there to match the control character (ASCII 7)
* The resulting output is leaving out the prefixed mode

If you want to replace it with this then I'll merge it in:
~~~
return word.replace(/(^|\\s|[@+~&%}]+)([#&][^ .,\007<>]+)$/i, (match, mPrefix, mChannel) => {
    let channel = _.escape(mChannel.trim());
    let prefix = _.escape(mPrefix);

    return `${prefix}<a class="u-link kiwi-channel" data-channel-name="${channel}">${channel}</a>`;
});
~~~

prawnsalad:
I wanted to get this fix in place so I've aded it here https://github.com/kiwiirc/kiwiirc/commit/311686ce3f503c036c7152fe85105eb5d5bafd6a

Thanks for the PR o/

PR #169, opened 2018-01-08
Styling changes to Kiwi 'Custom Server' Welcome screen

AllanOcelot:
I've changed the location to a more central location and adjusted the spacing of all elements to make everything tighter. This is just me getting used to working within the KiiwIRC repo and following it's coding standards etc.

Open to styling discussions and opinions that others may have.


prawnsalad:
I haven't tried it yet but a few notes:
* 4 space indentation in the CSS
* Components should not style things outside of its scope. ie. Welcome.vue should not style `body`
* Styling in the vue files should be bare minimal. Just enough to make it work, as the real theming (anything that a theme will msot likely change) should be in the theme files.

src/components/ContainerHeader.vue
```diff
@@ -13,8 +13,8 @@
             </div>
             <div class="kiwi-header-name">{{buffer.name}}</div>
             <div v-if="isJoined" class="kiwi-header-topic"></div>
-            <div v-if="!isJoined && isConnected" class="kiwi-header-notjoined">
-                <a @click="joinCurrentBuffer" class="u-link">{{$t('container_join')}}</a>
+            <div v-if="isJoined && isConnected" class="kiwi-header-notjoined">
+                <a @click="joinCurrentBuffer" class="u-link join-channel-button">{{$t('container_join')}}</a>
```
prawnsalad: note: class names like this should be prefixed with what section it's in. So here it should be `.kiwi-header-join-channel`

Issue #120, opened 2017-10-28
Hide/show unread message counter per channel


prawnsalad:

AllanOcelot:
This does not seem to have been implemented yet

ItsOnlyBinary:
This has been implemented on a per channel basis (it can be found in the notification settings)

PR #182, opened 2018-01-17
add option to show/hide unread message count to BufferSettings.vue

ItsOnlyBinary:
#120

### DIFF
--- a/src/components/BufferSettings.vue
+++ b/src/components/BufferSettings.vue
@@ -20,6 +20,10 @@
                     <span>{{$t('settings_notify_mute')}}</span>
                     <input type="checkbox" v-model="settingMuteSound">
                 </label><br />
+                <label>
+                    <span>{{$t('settings_show_message_counts')}}</span>
+                    <input type="checkbox" v-model="settingHideCount">
+                </label><br />
             </form>
         </div>
     </div>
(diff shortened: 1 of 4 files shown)


src/components/BufferSettings.vue
```diff
@@ -52,6 +56,14 @@ export default {
                 return this.buffer.setting('mute_sound', val);
             },
         },
+        settingHideCount: {
+            get: function getSettingAlertOn() {
+                return this.buffer.setting('hide_message_count');
+            },
+            set: function setSettingAlertOn(val) {
+                return this.buffer.setting('hide_message_count', val);
+            },
```
prawnsalad: These function names do not match the setting they're dealing with

src/res/locales/app.dev.po
```diff
@@ -233,6 +233,10 @@ msgstr "Never"
 msgid "settings_notify_mute"
 msgstr "Mute sounds"
 
+#: Hide unread message count
+msgid "settings_notify_count"
```
prawnsalad: Since the message counts are not notifications of sorts, thise should probably be renamed to `settings_show_message_counts`

src/components/StateBrowserNetwork.vue
```diff
@@ -103,6 +103,9 @@ export default {
         BufferSettings,
     },
     methods: {
+        showCount: function showCount(buffer) {
```
prawnsalad: This should be more specific and be named `showMessageCounts`

PR #187, opened 2018-01-18
make (channel/query) close button go to first available channel

ItsOnlyBinary:
#9

### DIFF
--- a/src/libs/state.js
+++ b/src/libs/state.js
@@ -745,8 +745,9 @@ const state = new Vue({
             }
 
             if (isActiveBuffer) {
-                let target = (network.buffers.length > 1) ?
-                    network.buffers[1].name : network.serverBuffer().name;
+                let target = network.buffers.length > 1 ?
+                    network.buffers[1].name :
+                    network.serverBuffer().name;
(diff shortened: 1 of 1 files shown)


src/libs/state.js
```diff
@@ -684,7 +684,9 @@ const state = new Vue({
             }
 
             if (isActiveBuffer) {
-                this.setActiveBuffer(network.id, network.serverBuffer().name);
+                let target = (network.buffers.length > 1) ?
+                    network.buffers[1].name : network.serverBuffer().name;
```
prawnsalad: This ternary should use the existing format used throughout the code:
~~~
let target = expr ?
    val1 :
    val2;
~~~

ItsOnlyBinary:
While this works and is much better than the current situation, it appears old client behaves slightly differently.
on the old client it appears to switch to the most recently visible channel.

CoryChaplin:
@ItsOnlyBinary I think this should go to previous open window (with a store of a history we roll down) like it was un kiwi v1.

ItsOnlyBinary:
@CoryChaplin see https://github.com/kiwiirc/kiwiirc/pull/216 ;)

PR #191, opened 2018-01-18
Channel Options Styling

AllanOcelot:
First commit is sorting out git issues with PR, simply includes style changes for the top 'Channel Options' button.

### DIFF
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -264,4 +264,10 @@ export default {
   to   { height: 100%; }
 }
 
+@media screen and (max-width: 600px){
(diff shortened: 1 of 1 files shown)


AllanOcelot:
@prawnsalad  - any notes / changes you would like making to this PR?

prawnsalad:
The close channel options text now has the arrow pointing left when closing actually slides it to the right. That's confusing and should probably be reverted to how it was.

Likewise with the all caps text on channel options text. That should be reverted back to lowercase as it shouldn't be trying to steal attention from the rest of the app.

Lastly, 4 spaces indentation throughout please :)

PR #208, opened 2018-01-23
Not connected display

AllanOcelot:
Created and styled a disconnect section, makes it far clearer to the user what is going on and presents them with the two options.

Note: It's trying to push the file  src/components/SelfUser.vue - I believe this to be due to Line endings on my windows machine. Anyone who can shed some light on this will have my thanks.


AllanOcelot:
Cleaning up git, will repush shortly.

PR #215, opened 2018-01-26
add focus directive

ItsOnlyBinary:
not perfect but very light

#89

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -81,6 +81,22 @@ Vue.directive('rawElement', {
     },
 });
 
+// Register a global custom directive called `v-focus`
+Vue.directive('focus', {
+    // When the bound element is inserted into the DOM...
+    inserted(el) {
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -81,6 +81,14 @@ Vue.directive('rawElement', {
     },
 });
 
+// Register a global custom directive called `v-focus`
+Vue.directive('focus', {
+    // When the bound element is inserted into the DOM...
+    inserted(el) {
+        // Focus the first input element
+        el.getElementsByTagName('input')[0].focus();
```
prawnsalad: This should check if `el.getElementsByTagName('input')[0]` actually exists before it calls `focus()`

prawnsalad:
To make this work with plain inputs as well as text-input you could check for what type of element `el` is. input = just call `focus()`. Anything else, do as you currently do.

PR #220, opened 2018-01-31
improved language detection

ItsOnlyBinary:


src/main.js
```diff
@@ -224,21 +224,27 @@ function initLocales() {
...
-            }
-        });
+        preferredLangs = _.merge([defaultLang], preferredLangs);
     }
+
+    _.forEach(preferredLangs, (lang) => {
+        if (_.includes(AvailableLocales.locales, lang.toLowerCase())) {
+            i18next.changeLanguage(lang, (err, t) => {
+                if (err) {
+                    i18next.changeLanguage('en-us');
```
CoryChaplin: This returns before your loop has finished, it needs to give a chance for all languages to match before you fall back to default.
Other approach, set this by default and override with the loop.

PR #227, opened 2018-02-04
Add a filter in the user list

DavidVerdier:
The filter function of users on big channels is util.


prawnsalad:
Nice nice. I've tweaked some of the styling so it's easier to work with in future and now merged. Thanks!

PR #229, opened 2018-02-05
Update fixed Welcome.vue

YanickFrehner:
fixed autoConnect when no captcha is set.


prawnsalad:
Good catch. Thanks o/

PR #234, opened 2018-02-06
Update app.bg-BG.po

Indjov:


prawnsalad:
This is set by the crowdin.com translation manager. I can't find a way to change this author automatically at the moment but I've asked them if they can get this implemented.

PR #296, opened 2018-04-13
Added support for iframes in the media viewer.

srmcgann:


prawnsalad:
This needs to be using the vuejs methods for managing the DOM as it creates the elements internally using a virtual DOM. Instead of the `createElement` calls, something like this in the template:

~~~javascript
<a
    v-if="isIframe"
    v-bind:href="url"
    class="embedly-card"
    :data-card-key="embedlyKey"
    data-card-chrome="0"
    data-card-controls="0"
    data-card-recommend="0"
>{{$t('media_loading', {url: url})}}</a>
<iframe
    v-else
    class="kiwi-mediaviewer-iframe"
    :src="url"></iframe>
~~~

Notice the `v-if` and `v-else` attributes. https://vuejs.org/v2/guide/#Declarative-Rendering

Then the iframe URL should not be in the `state.settings` object as it's local to this component only. So adding `isIframe: false` to the `data()` function and using that value to toggle between the embedly and iframe elements. The DOM will auto update itself when you change the `isIframe` value.

PR #306, opened 2018-04-25
Hotfix for Issue 299 and others.

AllanOcelot:
- Fixed issues with timestamp on compact running into text
- Fixed messages sometimes overrunning the messagelist / compact mobile
- Fixed connection messages not being aligned on mobile / compact view

### DIFF
--- a/src/components/MessageListMessageCompact.vue
+++ b/src/components/MessageListMessageCompact.vue
@@ -123,6 +123,7 @@ export default {
     display: inline-block;
     float: right;
     font-size: 12px;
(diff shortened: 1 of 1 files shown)


src/components/MessageListMessageCompact.vue
```diff
@@ -123,13 +123,15 @@ export default {
     display: inline-block;
     float: right;
     font-size: 12px;
+    line-height: 22px;
```
prawnsalad: Would this be better in em's so that themes can override the font-size easier?

Issue #57, opened 2017-07-24
Disable colour picker UI by default


prawnsalad:
Removing it from the UI prevents random people trying things out from annoying everyone else but keep the ctrl+k hotkey always enabled for people that know about it.

AllanOcelot:
#1000

prawnsalad:
yes

PR #308, opened 2018-04-26
Disable colour picker UI by default - Issue #57

AllanOcelot:
This addresses long standing item #57 - and is hopefully the start of me adding more user options in future 😁  .

- Added in option to display the colour picker, by default the colour picker will be hidden
- Added in translation string for new option


src/libs/state.js
```diff
@@ -69,6 +69,7 @@ const stateObj = {
             state_key: 'kiwi-state',
         },
         showAutocomplete: true,
+        showColour: false,
```
prawnsalad: Shouldn't this be `showColourPicker`?

src/components/ControlInput.vue
```diff
@@ -133,6 +134,10 @@ export default {
                 this.active_tool = tool;
             }
         },
+        //Check the users settings to see if they have turned on the colour picker
+        displayColourPicker: function displayColourPicker() {
+           this.display_colour_picker = state.setting('showColourPicker');
+        },
```
prawnsalad: This `displayColourPicker()` function would be better suited in the computed section of the component and the `display_colour_picker` data property removed. eg:
~~~javascript
shouldDisplayColourPicker() {
    return state.setting('showColourPicker');
}
~~~

You would then use it like `v-if="shouldDisplayColourPicker"`

prawnsalad:
The app.en-US.po translation file should not be modified. The translations.kiwiirc.com website handles all that for us based on the app.dev.po file.

AllanOcelot:
Outdated with the new version of Kiwi.

Issue #325, opened 2018-05-09
Nickname Mouse Over issue on Safari (osx) 


H7-25:
Using safari on osx, i've this issue : 

https://www.youtube.com/watch?v=ZstyL3-Mw5A

It look be related to  .kiwi-messagelist-nick overflow css , disabling  this in the inspector  up&down scrolling on mouse over look disappear

AllanOcelot:
@H7-25  - I'll look into this

AllanOcelot:
Hi @H7-25  - please can you check if the above commits (now in master) have resolved this issue?

H7-25:
I confirm, it fix the issue.

H7-25:
The issue is back again ...

AllanOcelot:
This does not appear to be an issue anymore, tested in Safari with master, using 'traditional' layout.

I cannot replicate, so am assuming a previous PR has addressed the issue :)

PR #327, opened 2018-05-09
Remove overflow visible from Nick on hover -  Issue #325 fix

AllanOcelot:
- Removed the overflow ; visible property from the MessageList component. Issue #325 (username spacing on hover, safari only) seems to be caused by this. I can see no negative side effect of removing this on other browsers, as nicknames that are very long are not fully shown, even on hover, at present.

#325  Fix.

### DIFF
--- a/src/components/MessageListMessageCompact.vue
+++ b/src/components/MessageListMessageCompact.vue
@@ -112,12 +112,12 @@ export default {
     position: absolute;
 }
 
-.kiwi-messagelist-message--compact.kiwi-messagelist-message-nick .kiwi-messagelist-time {
-    margin-right: 10px;
+.kiwi-messagelist-message--compact .kiwi-messagelist-nick:hover {
+    width: auto;
 }
 
-.kiwi-messagelist-message--compact .kiwi-messagelist-message-nick {
-    margin-right: 14px;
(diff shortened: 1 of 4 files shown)


prawnsalad:
Rather than removing this style which lets people actually see the full nick, lets try also setting `width:auto;` and setting the background to the same as the message hover background. This should make the nick be displayed in full and overlap the message so the text is not clashing

src/components/MessageListMessageCompact.vue
```diff
@@ -112,12 +112,12 @@ export default {
     position: absolute;
 }
 
-.kiwi-messagelist-message--compact.kiwi-messagelist-message-nick .kiwi-messagelist-time {
-    margin-right: 10px;
+.kiwi-messagelist-message--compact .kiwi-messagelist-nick:hover {
+    width: auto;
```
prawnsalad: Is this style not redundant as it's already set in MessageList.vue?

AllanOcelot: Compact view overrides width default for the Nick - and sets a width on it, we then over ride that set width with auto.

ModernLayout uses the default width auto and doesn't over ride it however

PR #342, opened 2018-05-15
Allow the nicklist to be pinned by default if the screen is wide enough

ItsOnlyBinary:


CoryChaplin:
Shouldn't this deserve an option in the config file ?

prawnsalad:
@CoryChaplin `"sidebarPinned":true` in the config file would auto pin the sidebar

Issue #364, opened 2018-05-26
Nick hovering on the dark theme has colour glitches


prawnsalad:
![darknickhover](https://user-images.githubusercontent.com/725880/40577856-79a43acc-6103-11e8-9ac0-ce46a3cf4348.gif)

PR #365, opened 2018-05-26
Fix username position change on hover - #364

AllanOcelot:
- Min width has been added to usernicks, to ensure when width auto is in effect on hover, the names stay in the same position

- Changed all form inputs, titles and labels to use pure white (#fff) rather than a light grey, as it improves legability


AllanOcelot:
- Another note, background colours of hovered messages have been changed for legibility also

PR #366, opened 2018-05-26
 Add in new MessageList Component, to display messages as text

AllanOcelot:
This PR deals with a few changes to Kiwi, most notably, it's my first draft of the 'Text' dispaly for MessageList. MessageListText is meant to display similar to modern chat rooms on video streaming websites.

- Added in new MessageListText component
- Select components have now been styled ( yay ! )
- Completely rewrote how the layout is displayed in the app settings. Rather than having an enable, disable option for Traditional, we now all users to select from a drop down list their preferred layout.

Note: This does not include translation strings for the new select. These will come in a separate PR.

### DIFF
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -39,8 +39,17 @@
                     <div class="kiwi-appsettings-block">
                         <h3>{{$t('settings_messages_title')}}</h3>
                         <div class="kiwi-appsettings-section kiwi-appsettings-messages">
+                            <label for="kiwi-select-layout">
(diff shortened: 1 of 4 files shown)


src/components/AppSettings.vue
```diff
@@ -221,6 +215,18 @@ export default {
                 this.$watch('customThemeUrl', watchCustomThemeUrl),
             ];
         },
+        settingMessageLayout: function settingMessageLayout(settingSelectedLayout) {
+            console.log('hi');
```
prawnsalad: hi.

PR #380, opened 2018-06-07
remember buffers by default if personal startup is used

ItsOnlyBinary:
#373


prawnsalad:
`main.js` shouldn't have any startup screens hardcoded as that's a slippery slope to spaghetti code. The `state.persist.loadStateIfExists()` line here could be added direct in the startup screens init function instead so that it's self contained.

Issue #381, opened 2018-06-07
Errors when deleting the last network in personal view


ItsOnlyBinary:
state.js:860
state.js:1248

https://imgur.com/a/Ztvn43w

PR #382, opened 2018-06-07
Ensure the network still exists before adding new messages to the buffer

ItsOnlyBinary:
#381 

this error is caused by the socket close event which IrcClient tries to add a message for

### DIFF
--- a/src/libs/state.js
+++ b/src/libs/state.js
@@ -858,7 +858,8 @@ const state = new Vue({
             let network = buffer.getNetwork();
             let isNewMessage = message.time >= buffer.last_read;
             let isHighlight = !network ?
-                false : Misc.mentionsNick(bufferMessage.message, network.ircClient.user.nick);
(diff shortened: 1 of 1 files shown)


src/libs/state.js
```diff
@@ -857,7 +857,8 @@ const state = new Vue({
 
             let network = buffer.getNetwork();
             let isNewMessage = message.time >= buffer.last_read;
-            let isHighlight = Misc.mentionsNick(bufferMessage.message, network.ircClient.user.nick);
+            let isHighlight = !network ?
+                false : Misc.mentionsNick(bufferMessage.message, network.ircClient.user.nick);
```
prawnsalad: For consistency this should be structured as:
~~~javascript
            let isHighlight = !network ?
                false :
                Misc.mentionsNick(bufferMessage.message, network.ircClient.user.nick);
~~~

Issue #391, opened 2018-06-12
self-popup hard to close


strarsis:
When clicking ones own name in kiwirc, a self-popup appears, this can also happen accidentally.
It was very hard for me figuring out how to close it (click again on the name to make it collapse)
and the popup actually covers the conversation text behind it.

There is no typical cross icon at the corner for example.
In the end I had to remove the DOM elements using browser inspector 
(because I didn't want to reload the whole page and restart the client).

AllanOcelot:
#393 should resolve this once merged in :)

ItsOnlyBinary:
The fix for this has been merged into master. Thanks for the report

PR #393, opened 2018-06-12
Make showing and hiding SelfUser.vue cleaner

ItsOnlyBinary:
#391

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -1,13 +1,13 @@
 <template>
     <div class="kiwi-controlinput">
         <div class="kiwi-controlinput-selfuser" :class="{'kiwi-controlinput-selfuser--open': selfuser_open}">
-            <self-user :network="buffer.getNetwork()" v-if="selfuser_open && networkState==='connected'"></self-user>
+            <self-user :network="buffer.getNetwork()" v-if="selfuser_open && networkState==='connected'" :parent=this></self-user>
         </div>
 
         <div class="kiwi-controlinput-inner">
             <div v-if="currentNick" class="kiwi-controlinput-user" @click="toggleSelfUser">
                 <span class="kiwi-controlinput-user-nick">{{ currentNick }}</span>
-                <i class="fa fa-caret-up" aria-hidden="true"></i>
+                <i :class="[selfuser_open ? 'fa-caret-down' : 'fa-caret-up']" class="fa" aria-hidden="true"></i>
             </div>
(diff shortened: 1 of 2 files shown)


src/components/ControlInput.vue
```diff
@@ -399,6 +399,10 @@ export default {
     box-sizing: border-box;
 }
 
+.kiwi-controlinput .kiwi-controlinput-inner i {
```
prawnsalad: This can just be `.kiwi-controlinput-inner i`

src/components/SelfUser.vue
```diff
@@ -1,5 +1,8 @@
 <template>
     <div class="kiwi-selfuser">
+        <div class="close-icon" @click="hide">
```
prawnsalad: `close-icon` should be named correctly by prefixing the component name at the start. Eg. `kiwi-selfuser-close`

src/components/SelfUser.vue
```diff
@@ -31,13 +34,16 @@ export default {
                 str = '+' + str;
             }
 
-            return str;
+            return 'User modes: ' + str;
```
prawnsalad: The 'User modes: ' text isn't needed here as it's obvious what it is already while keeping it less cluttered

src/components/SelfUser.vue
```diff
@@ -31,13 +34,16 @@ export default {
...
         },
         netUser() {
             return this.network.ircClient.user;
         },
     },
     methods: {
+        hide: function hide() {
+            this.parent.selfuser_open = false;
```
prawnsalad: Child components should never interact with its parent directly and passing it in as a prop is just a long about way of doing that. It's better to emit a `close` event and use `@close="selfuser_open=false"` on the parent instead

src/components/SelfUser.vue
```diff
@@ -54,6 +60,19 @@ export default {
...
+    right: 0;
+    top: 0;
+    color: #fff;
+    cursor: pointer;
+    font-size: 0.8em;
+    padding: 0.2em 0.4em;
+    border-radius: 0 0 0 0.4em;
+    transition: all 0.3s;
+    background-color: #fc6262;
+}
```
prawnsalad: Colour related styles should be in the theme files, not in the core

src/components/SelfUser.vue
ItsOnlyBinary: im not convinced this is clear how it originally was for me it was just a line with +cx but no explanation of that +cx is

ItsOnlyBinary: image

PR #396, opened 2018-06-14
Globally style close icons

AllanOcelot:
Close icons are now globally styled with the class 'kiwi-close-icon'. This is meant to be merged with #393 to ensure it is styled correctly

### DIFF
--- a/src/components/StateBrowser.vue
+++ b/src/components/StateBrowser.vue
@@ -17,7 +17,7 @@
             <div v-if="is_usermenu_open" class="kiwi-statebrowser-usermenu-body">
                 <p> {{$t('state_remembered')}} </p>
                 <a @click="clickForget" class="u-link">{{$t('state_forget')}}</a>
-                <div class="close-icon" @click="is_usermenu_open=false">
+                <div class="kiwi-close-icon" @click="is_usermenu_open=false">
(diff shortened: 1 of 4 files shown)


prawnsalad:
Shouldn't the structural styles for the close button be in the core, not the themes? Ie. border-border-radius/position/top/right/cursor. Also, you have 2 border-radius entries,

~~~css
+    background-color: #fc6262;
+    color: #fff;
+    border-radius: 0 0 0 4px;
+    position: absolute;
+    right: 0;
+    top: 0;
+    cursor: pointer;
+    padding: 0.2em 0.4em;
+    border-radius: 0 0 0 0.4em;
~~~

AllanOcelot:
@prawnsalad  - that's true but I feel if a component is being styled exactly the same ( we want all close icons across the board to look / feel the same) that they should share the structual css.

prawnsalad:
@AllanOcelot: right, that's what I mean. We do have a global stylesheet https://github.com/kiwiirc/kiwiirc/blob/master/src/res/globalStyle.css

AllanOcelot:
You sir, are correct aha.

I forgot about that, Got carried away with the themes this week. I'll add this in now.

PR #400, opened 2018-06-16
fix build by transpiling node_modules

vith:
excluded a couple of modules that have errors when passed through babel


vith:
Note: the motivation for these changes was build failures, and later, es6 library code leaking into the webpack bundles when symlinking the master branch of `irc-framework` into `./node_modules/` (i.e. with `yarn link`).

I'm not sure if it would be wiser to selectively transpile the problematic dependencies (possibly just `runes` at this point) instead of transpiling them all by default and using exclusions.

PR #401, opened 2018-06-18
When adding a network also add it to the recent buffers list

ItsOnlyBinary:
This is because you can end up on the kiwi-container-empty screen when you close the only item in the recently used buffers.


ItsOnlyBinary:
The alternate to this would be to just try and find a random buffer or network

prawnsalad:
This PR doesn't make sense since you're not viewing the server tab, it's only adding it. If there are no items left in the buffer list then selecting the server tab on either active network, or the first network in that order, would make more logical sense.

PR #402, opened 2018-06-18
Create a default FallbackComponent

ItsOnlyBinary:
This hopefully will solve the issue of the input being shown when there is no buffer due to kiwi-container-empty being shown

kiwi-container-empty has now been moved to FallbackComponent.

**Due to limited time testing of this commit has been limited**


ItsOnlyBinary:
To see the issue this is fixing connect to irc with one channel then close that one channel and try to type something into input box

ItsOnlyBinary:
could not find any situations where this is still required

PR #407, opened 2018-06-21
revised connecting animation - added duration, easing out

srmcgann:


prawnsalad:
Linting of .vue is now working again which is now catching a few things in this branch.

I've also tidied up a few weird sizing+positioning things to keep things clearer. The last thing remaining here is to clean up a sizing oddity. From what I can see, it's rendering at a huge px size (1000px+), but appears really small, and then uses 150% h+w to increase its size? I expect this is why when it's set to size 150x150px in the MessageList, the actual space used is wider than that with an empty space to the right of the rendered canvas.

image

PR #425, opened 2018-07-02
Add whowas support

ItsOnlyBinary:
I'm not totally happy with the error output. but wasn't sure how to make it better.


prawnsalad:
Since this is always in direct response to the /whowas command (unlike whois which may be part of clicking a button in the userbox) then this code would be ebtter suited in the Inputhandler.js file which adds all the user commands, simialr to: https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/InputHandler.js#L457

ItsOnlyBinary:
Requires https://github.com/kiwiirc/irc-framework/pull/163

Issue #31, opened 2017-04-29
Input field contents cleared after visiting options


mestaritonttu:
1. Type something to the input field
2. Click the options
3. Close the options

Your typed text is gone.

PR #442, opened 2018-07-11
Remember IrcInput either commonly or per buffer

ItsOnlyBinary:
#31

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -37,6 +37,7 @@
                         :placeholder="$t('input_placeholder')"
                         class="kiwi-controlinput-input"
                         wrap="off"
+                        @input="inputUpdate"
                         @keydown="inputKeyDown($event)"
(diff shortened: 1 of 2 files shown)


src/components/ControlInput.vue
```diff
@@ -148,7 +154,22 @@ export default {
...
+        inputUpdate(val) {
+            if (state.getSetting('settings.buffers.shared_input')) {
+                state.ui.current_input = val;
+            } else {
+                this.buffer.current_input = val;
+            }
+        },
+        inputRestore() {
+            let currentInput = state.getSetting('settings.buffers.shared_input') ?
+                state.ui.current_input : this.buffer.current_input;
```
prawnsalad: Ternary should be in the form:
~~~javascript
let currentInput = state.getSetting('settings.buffers.shared_input') ?
    state.ui.current_input :
    this.buffer.current_input;
~~~

PR #443, opened 2018-07-12
fix translations for whois

ItsOnlyBinary:


prawnsalad:
These need setting on crowdin as they are managed there, not directly editing these .po files

PR #451, opened 2018-07-13
EXPERIMENTAL: Giphy integration 

AllanOcelot:
Adding in Giphy to Kiwi will give us the same level of functionality that many of our users have come to expect from Kiwi.

This is an experimental build, currently it is unstyled and there are plenty of bugs whilst we try to decide on the best method for inserting the gifs into the messageList component. I belive we are parsing for HTML tags in messages, and as such, it is removing the gifs when a user submits their gif.

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -51,7 +51,7 @@
                     <i class="fa fa-smile-o" aria-hidden="true"/>
                 </a>
                 <a class="kiwi-controlinput-tool" @click.prevent="onToolClickGiphy">
-                    <i class="fa fa-picture-o" aria-hidden="true"></i>
+                    <i class="fa fa-picture-o" aria-hidden="true"/>
                 </a>
                 <div
                     v-rawElement="plugin.el"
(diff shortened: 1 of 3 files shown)


AllanOcelot:
Update:

So , we need to add in some kind of sniffing for .gif image links, that will display the gif inline in Kiwi.

I am however, aware our users are pretty split, so I would not want to display gifs unless the end user wants them displayed. We could do this with another user option: 'Enable inline image's' or something similar. I can provide and wire up the option but I am out of my depth when it comes to displaying the images inline in the messageList.

Anyone who feels they can help out, please feel free to point me in the right direction ❤️

Shillos:
Would be nice to see this feature in soon!

AllanOcelot:
Hi @Shillos  -  we are still working on it, hopefully the bugs will be ironed out soon :)

Shillos:
Very much looking forward to this! Thanks!

Shillos:
Any luck with this one?!

AllanOcelot:
@Shillos  - Our very own @ItsOnlyBinary  managed to take my code and made a working plugin that can added to your Kiwi install.  ( https://github.com/ItsOnlyBinary/my-kiwiirc-plugins ).

Shillos:
Oh that's great! I was wondering if this could be released as a plugin. Thanks!

AllanOcelot:
@ItsOnlyBinary  has created a fully featured plugin that achieves this. There's no need for this PR anymore.

PR #452, opened 2018-07-13
Mouse events on firefox do not have the e.srcElement using e.target instead

ItsOnlyBinary:
https://developer.mozilla.org/en-US/docs/Web/API/Event/target#Compatibility_notes


prawnsalad:
My bad, I missed this PR while committing the same fix here https://github.com/kiwiirc/kiwiirc/commit/3e00cef815d95130fdb9fc36656dc307238dd7f1

Issue #367, opened 2018-05-27
Breadcrumbs at the top of the sidebar


prawnsalad:
Specifically, nicklist > user box

AllanOcelot:
@prawnsalad  - Would it suffice to have static bread crumbs, i.e - just a text tag with 'Nicklist > User Details' under the sidebar title? This would be easy enough for me to add in?.

PR #458, opened 2018-07-18
Adding breadcrumbs to sidebar

AllanOcelot:
@prawnsalad  - is there a way to get the 'active' sidebar component. I'd like to have the breadcrumbs on the sidebar reactive to the current component, rather than having a different string at the top of each component :)


AllanOcelot:
For #367  :)

PR #461, opened 2018-07-19
find a buffer after removing network

ItsOnlyBinary:


prawnsalad:
This should mark if it's the active network, remove the network, and *then* call `openLastActiveBuffer()`. Otherwise some checks in `openLastActiveBuffer()` will just pick up the network that's about to be deleted in some cases

ItsOnlyBinary:
this is ready for merging

Issue #453, opened 2018-07-15
Advanced settings table similar to firefox about:config


prawnsalad:
To walk through all proeprties under state.settings, listing each key as dotted notation in a table. Make sure to be reading the overridden user settings from state.user_settings

AllanOcelot:
Design for this is here:
https://i.imgur.com/ObeTKQE.png

PR #466, opened 2018-07-21
Add advanced settings tab

ItsOnlyBinary:
#453

### DIFF
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -246,10 +246,10 @@ export default {
         settingDefaultKickReason: bindSetting('buffers.default_kick_reason'),
         settingAdvancedEnable: {
             get: function getSettingShowAdvancedTab() {
-                return state.getSetting('ui.show_advanced_tab');
+                return state.ui.show_advanced_tab;
             },
             set: function setSettingShowAdvancedTab(newVal) {
-                state.setSetting('ui.show_advanced_tab', newVal);
+                state.ui.show_advanced_tab = newVal;
             },
         },
         settingMessageLayout: {
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -128,7 +128,7 @@
                         <div class="kiwi-appsettings-section kiwi-appsettings-advanced-enable">
                             <div>
                                 <span style="font-weight: 600;">
-                                    {{ $t('settings_advanced_warning_title') }}
+                                    {{ $t('warning') }}
                                 </span>
                                 {{ $t('settings_advanced_warning') }}
                             </div>
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -421,6 +421,11 @@ export default {
     margin: 20px auto 20px auto;
 }
 
+.kiwi-appsettings-block.kiwi-appsettings-block-advanced {
+    max-width: inherit;
+    margin: 20px;
+}
+
 .kiwi-appsettings-block h3 {
     width: 100%;
     line-height: 45px;
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -122,6 +122,17 @@
                             </label>
                         </div>
                     </div>
+                    <div v-if="!state.setting('hide_advanced') && !settingAdvancedEnable"
+                         class="kiwi-appsettings-block">
+                        <h3>{{ $t('settings_advanced_title') }}</h3>
+                        <div class="kiwi-appsettings-section kiwi-appsettings-advanced-enable">
+                            <div><span>{{ $t('settings_advanced_warning_title') }}</span>
+                                {{ $t('settings_advanced_warning') }}
+                            </div>
+                            <a class="u-button" @click="settingAdvancedEnable = true"
+                            >{{ $t('settings_advanced_button') }}</a>
+                        </div>
+                    </div>
                 </tabbed-tab>
 
                 <tabbed-tab :header="$t('settings_aliases')">
@@ -132,6 +143,18 @@
                         </div>
                     </div>
                 </tabbed-tab>
+
+                <tabbed-tab
+                    v-if="settingAdvancedEnable"
+                    :header="$t('settings_advanced')">
+                    <div class="kiwi-appsettings-block kiwi-appsettings-block-advanced">
+                        <h3>{{ $t('settings_advanced') }}</h3>
+                        <div class="kiwi-appsettings-section kiwi-appsettings-advanced">
+                            <settings-advanced/>
+                        </div>
+                    </div>
+                </tabbed-tab>
+
                 <tabbed-tab v-for="item in pluginUiElements" :key="item.id" :header="item.title">
                     <div :is="item.component" v-bind="item.props"/>
                 </tabbed-tab>
@@ -146,6 +169,7 @@ import state from '@/libs/state';
 import ThemeManager from '@/libs/ThemeManager';
 import GlobalApi from '@/libs/GlobalApi';
 import SettingsAliases from './SettingsAliases';
+import SettingsAdvanced from './SettingsAdvanced';
 
 /**
  * Returns an object for a vuejs computated property on a state settings value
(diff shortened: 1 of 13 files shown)


AllanOcelot:
Very happy with this, looks good to me. 
My only suggestion would be to replace the 'advanced' title bar, and instead style the table header the same way - so the 'setting name, status etc' sit where 'Advanced' is now.

@ItsOnlyBinary  - Happy to take a look at this first thing monday if that sounds good?

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,164 @@
...
+                    <td>
+                        <input v-if="setting.type === 'boolean'"
+                               :checked="setting.val"
+                               type="checkbox"
+                               @click="updateSetting($event, setting.key)">
+                        <input v-else-if="setting.type === 'number'"
+                               :value="setting.val"
+                               class="u-input"
+                               type="number"
+                               @keydown="blurOnEnter($event)"
```
prawnsalad: This can be simplified with `@keydown.13="$event.target.blur()"`

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,164 @@
...
+
+export default {
+    data: function data() {
+        return {
+            ignoreKeys: ['emojis', 'themes', 'bnc', 'aliases', 'restricted',
+                'hide_advanced', 'windowTitle', 'startupOptions'],
+        };
+    },
+    computed: {
+        getSettings() {
```
prawnsalad: `getSettings` is superfluous as a computed property (as a method it would make sense). This should just be `settings`

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,164 @@
...
+                            type: typeof val,
+                            status: status,
+                        };
+                    }
+                } else if (typeof val === 'object') {
+                    this.buildTree(data, ourBase, object[key], status);
+                }
+            }
+        },
+        bindSetting(settingName) {
```
prawnsalad: Is `bindSetting()` used anywhere?

src/components/AppSettings.vue
```diff
@@ -208,6 +244,14 @@ export default {
...
         settingDefaultBanMask: bindSetting('buffers.default_ban_mask'),
         settingDefaultKickReason: bindSetting('buffers.default_kick_reason'),
+        settingAdvancedEnable: {
+            get: function getSettingShowAdvancedTab() {
+                return state.getSetting('ui.show_advanced_tab');
+            },
+            set: function setSettingShowAdvancedTab(newVal) {
+                state.setSetting('ui.show_advanced_tab', newVal);
+            },
+        },
```
prawnsalad: Is this not the same as `settingAdvancedEnable: state.ui.show_advanced_tab `?

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
+<template>
+    <div class="kiwi-settings-advanced">
+        <div class="kiwi-settings-advanced-notice">{{ $t('settings_advanced_header') }}</div>
+        <form class="u-form">
+            <div class="kiwi-settings-advanced-filter">
+                <div class="kiwi-settings-filter-container">
+                    <input id="filter" v-model="filterString"
```
prawnsalad: I can't see a reason why this input has an ID set. Also we don't use IDs on elements anywhere just on the off-chance they conflict with another element of the same ID at some point in future.

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+                           :placeholder="$t('settings_advanced_filter')"
+                           class="u-input">
+                    <i v-if="!filterString" class="fa fa-search" aria-hidden="true"/>
+                    <i v-else class="fa fa-times" aria-hidden="true" @click="filterString = ''"/>
+                </div>
+            </div>
+            <table class="u-table kiwi-settings-advanced-table" cellspacing="0">
+                <p v-if="filteredSettings.length === 0" class="kiwi-settings-advanced-noresult">
```
prawnsalad: A `<p>` element cannot be a child of a `<table>`.  This needs to be wrapped in a `<tr><td>` set of elements.

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+                    {{ $t('settings_advanced_noresult') }}
+                </p>
+                <tr v-for="setting in filteredSettings" v-else
+                    :key="setting.key"
+                    :style="{'font-weight': (setting.modified) ? 'bold' : 'normal' }">
```
prawnsalad: This would be better adding a CSS class modifier (eg. .kiwi-settings-advanced-row and then .kiwi-settings-advanced-row--modified as needed) so that all styling is in the style blocks at the bottom of the file, and also allows easier theming from theme files

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+                               type="number"
+                               @keydown.13="$event.target.blur()"
+                               @change="updateSetting($event, setting.key)"
+                               @blur="updateSetting($event, setting.key)">
+                        <input v-else
+                               :value="setting.val"
+                               :id="setting.key"
+                               class="u-input"
+                               @keydown.13="$event.target.blur()"
+                               @blur="updateSetting($event, setting.key)">
```
prawnsalad: All of these IDs need prefixing with something to prevent clashes with anything else on the page. Eg. a setting called `filter` would then clash with an element elsewhere on the page with the ID `filter`.

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+            ignoreKeys: ['emojis', 'themes', 'bnc', 'aliases', 'restricted',
+                'hide_advanced', 'windowTitle', 'startupOptions', 'plugins'],
+        };
+    },
+    computed: {
+        filteredSettings() {
+            let settings = this.settings;
+            let out = [];
+            Object.values(settings).forEach((value, index) => {
+                if (value.key.toLowerCase().includes(this.filterString.toLowerCase())) {
```
prawnsalad: We can get a performance improvement and easier code reading if you move this filterString property outside of the loop. `let filter = this.filterString.toLowerCase();` then just using that filter var.

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+    },
+    methods: {
+        resetValue(event, settingKey) {
+            let newVal = state.getSetting('settings.' + settingKey);
+            if (!newVal) {
+                newVal = null;
+            }
+            state.setting(settingKey, newVal);
+        },
+        blurOnEnter(event) {
```
prawnsalad: Is this used anywhere anymore?

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+                            <i class="fa fa-undo reset-icon" style="margin-left: 10px;"/>
+                        </a>
+                    </td>
+                    <td v-else />
+                    <td>
+                        <input v-if="setting.type === 'boolean'"
+                               :checked="setting.val"
+                               :id="setting.key"
+                               type="checkbox"
+                               @click="updateSetting($event, setting.key)">
```
prawnsalad: This won't pick up any changes that are not the result of a click. keyboard events, tap event, and probably others can also change the value.

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
+<template>
+    <div class="kiwi-settings-advanced">
+        <div class="kiwi-settings-advanced-notice">{{ $t('settings_advanced_header') }}</div>
+        <form class="u-form">
+            <div class="kiwi-settings-advanced-filter">
+                <div class="kiwi-settings-filter-container">
```
prawnsalad: This should be prefixed with `kiwi-settings-advanced-`

src/components/SettingsAdvanced.vue
```diff
@@ -0,0 +1,230 @@
...
+                }
+            });
+        },
+    },
+};
+</script>
+
+<style>
+
+.kiwi-appsettings-block-advanced {
```
prawnsalad: If this is styling something in the appsettings component, then this style should be in that component rather than here.

Issue #468, opened 2018-07-21
Support dotted notation in the config


prawnsalad:
Eg:
~~~json
{ "buffers.traffic_as_activity": false }
~~~

would be the same as:
~~~json
{
    "buffers": {
        "traffic_as_activity": false
    }
}
~~~

Dotted notation should also inherit its current position, so that the following are equal:
~~~json
{
    "buffers": {
        "sub.setting": false
    }
}

{
    "buffers": {
        "sub": {
            "setting": false
        }
    }
}
~~~

PR #471, opened 2018-07-24
Allow dot notation in config

ItsOnlyBinary:
#468


prawnsalad:
There are 2 parts to the loading of the config. The ConfigLoader which only deal with loading the config from different sources and applying and text replacements, then the applying of that config to the application https://github.com/kiwiirc/kiwiirc/blob/master/src/main.js#L171

You will need to put this dotted notation support alongside the applying of the config and not where it loads the config - these are two logically separate parts of the application which would be tested ( ...) seperately.

Issue #474, opened 2018-07-28
Channel names in the channel list are white on white background


prawnsalad:
Osprey theme. Needs checking on other themes too

AllanOcelot:
Looking into this one now :)

PR #480, opened 2018-08-03
Update: - Channel list styling

AllanOcelot:
Reworked the chanenl list (Channel Search) component to be more inline with how the new kiwi looks.
This PR also includes a fix for #474 .

### DIFF
--- a/src/components/ChannelList.vue
+++ b/src/components/ChannelList.vue
@@ -22,13 +22,6 @@
                 </div>
             </div>
             <table v-if="!isLoading && list.length > 0" :key="last_updated" width="100%">
-                <thead>
-                    <th class="kiwi-channellist-user-center">Users</th>
-                    <th>Channel name</th>
-                    <th>Channel topic</th>
-                    <th class="kiwi-channellist-user-center"
-                        style="min-width:100px;">Join Channel</th>
-                </thead>
                 <tbody>
                     <tr v-for="channel in paginated" :key="channel.channel">
                         <td class="kiwi-channellist-user-center">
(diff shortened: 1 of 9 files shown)


src/components/ChannelList.vue
```diff
@@ -1,46 +1,61 @@
...
-                <a @click="prevPage"><i class="fa fa-step-backward" aria-hidden="true"/></a>
-                {{ page + 1 }} / {{ maxPages + 1 }}
-                <a @click="nextPage"><i class="fa fa-step-forward" aria-hidden="true"/></a>
+    <div :class="{'kiwi-channellist-padding-top': !list.length}" class="kiwi-channellist">
+        <div class="kiwi-channellist-content-container">
+            <div class="kiwi-channellist-nav">
+                <form class="u-form kiwi-channellist-search" @submit.prevent>
+                    <input v-model="search" :placeholder="$t('do_search')" class="u-input" >
+                    <a
+                        :class="{
```
ItsOnlyBinary: put class on the same line as a, lint will then make you line everything under :class, should make it a little neater

ItsOnlyBinary: i say this then look into StateBrowserNetwork.vue and its all done your way. so maybe ignore me

prawnsalad:
What about simplifying this a bit by removing the titles as they were self explanatory, and the centre alignment for the channel name. Then also removing the channel modes from the topic string (that parts between the [ ] at the start).

The only weird part now I think is the join button that uses an icon commonly used for quit/exit actions.

image

Issue #478, opened 2018-08-03
Keyboard shortcut keys


prawnsalad:

prawnsalad:
`meta + [` = previous tab
`meta + ]` = next tab
`meta + b` = enable bold
`meta + b` = enable bold

ItsOnlyBinary:
meta + k is already used to bring up colour picker

PR #482, opened 2018-08-06
Keyboard shortcuts

ItsOnlyBinary:
#478


prawnsalad:
Looks mostly pretty good though a few observations:
- `BufferTools.js` should be `BufferTools.js`. Capitalised names are for constructors (ie. classes)
- `setBold` / `setItalic` / `setUnderline` looks like they should be `toggle*` instead? As it's toggling instead of always setting.

PR #486, opened 2018-08-08
Updating the userlist layout

AllanOcelot:
We have moved the filter to the top of the layout, added in a function to show/hide it and removed the 'users here' text string, leaving only the int visible.


AllanOcelot:
@prawnsalad  this needs reviewing again since the last changes, it should be ready for a merge with Master now 👍

Issue #503, opened 2018-08-14
Dark Theme: Search user list entry text for search to dark after PR#486


TheBeerbarian:
![image](https://user-images.githubusercontent.com/24723440/44123943-e07700c2-9fe7-11e8-9ddc-1b9a1618f8d5.png)

AllanOcelot:
Nice :)

TheBeerbarian:
I added a fix with PR #505 for just the dark theme.   If that was the
correct fix I could do the same for the other themes that needed a slightly
lighter text for the search input.

Cowan

TheBeerbarian:
Closed PR505 to new PR508 so would be from local branch. Made suggested review changes from PR505.

TheBeerbarian:
Fixed with PR 508.

PR #505, opened 2018-08-15
Add style tag to dark theme for issue #503

TheBeerbarian:


prawnsalad:
A better fix for this would be to add the following to https://github.com/kiwiirc/kiwiirc/blob/master/src/res/globalStyle.css#L18

~~~css
input,
textarea,
select {
    color: inherit;
}
~~~

TheBeerbarian:
Agreed on alternate suggestion. :)

PR #510, opened 2018-08-15
Invite support [depends on updated irc-framework]

srmcgann:


prawnsalad:
Now that a new irc-framework release is included, could you rebase this back onto master so we can be testing it out

Issue #498, opened 2018-08-14
Kiwi animation continues when attempting to join channels that are banned/key/invite


ItsOnlyBinary:
Leaves me feeling like im waiting for something to happen

PR #511, opened 2018-08-16
added criteria for displaying loading animation (addressing issue #498)

srmcgann:


prawnsalad:
This PR relied on specific language strings from the IRCd which could change at any point - and some non-english networks do just that.

I've added this commit instead which covers all channel errors https://github.com/kiwiirc/kiwiirc/commit/9a8b75279a0afaddbc5ce0dfb3370768b7d51ea8

PR #517, opened 2018-08-17
moved messageLayout setting to per buffer, shortened function syntaxes, changed `state.` to `this.$state.`

srmcgann:


src/components/MessageList.vue
```diff
@@ -71,6 +71,7 @@
 
 <script>
 
+import Logger from '@/libs/Logger';
```
prawnsalad: Just to be consistent, this should be moved down a few lines to the other local imports. ie. after the strftime module import line

PR #542, opened 2018-09-04
add bold, italic, underline to TextStyle.vue

ItsOnlyBinary:
I also tweaked the css names to more match the file name


prawnsalad:
This could now do with `user-select: none;` adding to the items so that the text content doesn't get highlighted when tapping on them

prawnsalad:
This PR seems to have also changed the default colour palette for some reason, which now some of them are difficult to differentiate from others and look generally harsh. Can you revert the colour changes and just keep this to the main topic, bold, italic, underline

ItsOnlyBinary:
I changed the colour codes to better match the mirc colours, which seem to be the standard for irc

ItsOnlyBinary:
Colour changes removed via rebase + forcepush

ItsOnlyBinary:
This needs revisiting, its been reported that its not possible to set multiple formats at the same time (eg Bold & italic)

AllanOcelot:
@ItsOnlyBinary  - Going to quickly dive into this PR today whilst I look at the colour Picker we currently have 😄

AllanOcelot:
@ItsOnlyBinary  - please ignore my linting errors with last commit, will be fixed shortly.

PR #543, opened 2018-09-05
added icon to collapse plugin UI. (Closes #539)

### LINKED ISSUE #539: Smiley and Textformatting button
- An option to hide smiley and text formatting button should be needed . 
- On mobile view they also should need a better solution.
- Smiley button on mobile device should be hidden by default, mobile device have his own emoji keyboard

srmcgann:


prawnsalad:
This is pretty nice. To take it a step further, add the `.kiwi-controlinput-tools-container` element in a vue transition so that we can animate it out of display

PR #545, opened 2018-09-06
Show an error if /join is used without a channel

### LINKED ISSUE #544: /join command with no argument in a parted channel buffer joins "#" instead
This should probably do nothing for non-channel buffers and issue a join to the current channel for channel buffers (i.e. when you parted, were kicked, etc., but didn't close a buffer).

ItsOnlyBinary:
fixes #544

### DIFF
--- a/src/libs/InputHandler.js
+++ b/src/libs/InputHandler.js
@@ -205,9 +205,17 @@ inputCommands.join = function inputCommandJoin(event, command, line) {
     let network = this.state.getActiveNetwork();
     let bufferObjs = Misc.extractBuffers(line);
 
-    // report an error if the user tries to join without specifying the channel
+    // handle join without any buffers specified
     if (bufferObjs.length === 0) {
         let buffer = this.state.getActiveBuffer();
+
+        // join the active channel if its not joined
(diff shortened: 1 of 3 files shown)


src/libs/InputHandler.js
```diff
@@ -207,10 +207,15 @@ inputCommands.join = function inputCommandJoin(event, command, line) {
 
     // Only switch to the first channel we join if multiple are being joined
     let hasSwitchedActiveBuffer = false;
-    bufferObjs.forEach((bufferObj) => {
-        // /join 0 parts all channels and is only ever used to troll IRC newbies.
```
ItsOnlyBinary: I removed this because # is appended to the 0 so does not part all channels

prawnsalad:
The root issue here is Misc.extractBuffers() converting empty buffer names to a `#`.
Adding the following here will solve this https://github.com/kiwiirc/kiwiirc/blob/master/src/helpers/Misc.js#L24
~~~javascript
        if (!bufferName.trim()) {
            return;
        }
~~~

Then back in InputHandler.js you can check if `bufferObjs.length === 0` to show the warning message

PR #549, opened 2018-09-09
Disable channel mode switches when user is not an op

### LINKED ISSUE #176: Channel modes + settings readonly when no chan op modes

ItsOnlyBinary:
closes #176

@AllanOcelot this could do with some styling so the user can easily see they are disabled

### DIFF
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -27,6 +27,15 @@
 
                     <tabbed-view>
                         <tabbed-tab :header="$t('settings')" :focus="true">
+                          <h3>{{ $t('channel_settings') }}
+                              <span v-if="!areWeAnOp()"
+                                    class="kiwi-channelsettings-disabled">
+                                    {{ $t('mods_only') }}
+                                    <i class="fa fa-lock" aria-hidden="true"></i>
+                              </span>
+                          </h3>
+                          <hr>
+                          <channel-info :buffer="buffer"/>
                             <div class="kiwi-sidebar-settings">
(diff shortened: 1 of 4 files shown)


ItsOnlyBinary:
^^^ rebased

prawnsalad:
Making all the content 50% transparent to show that they're disabled is just making it difficult to see here. How about instead of the thick outlines that the editable inputs have, they are removed so they're  more like labels instead of editable areas?

AllanOcelot:
@prawnsalad  and @ItsOnlyBinary  this could do with another look  - I've tried instead to just put a notice at the top of the channel settings, simple and effective :)

ItsOnlyBinary:
Don't think i like that either :( what about just hiding the options all together if they are not op.

that way at least it would make the `Your settings` more visible

src/components/Sidebar.vue
```diff
@@ -27,10 +27,15 @@
 
                     <tabbed-view>
                         <tabbed-tab :header="$t('settings')" :focus="true">
-                            <h3>{{ $t('channel_settings') }}</h3>
-                            <hr>
-                            <channel-info :buffer="buffer"/>
-
+                          <h3>{{ $t('channel_settings') }}
+                              <span v-if="!areWeAnOp()"
+                                    class="kiwi-channelsettings-disabled">
```
ItsOnlyBinary: newline the `if` and `>`

AllanOcelot:
@ItsOnlyBinary  @prawnsalad  did we come to a conclusion on this PR?

ItsOnlyBinary:
rebased

PR #558, opened 2018-09-17
Fix channel linkify

ItsOnlyBinary:


prawnsalad:
Channels can contain characters `;:` so we should only be trimming them from the end, not skipping them entirely.

Some quick mini tests with a modified regex:
~~~javascript
tests=[
	['#channel', '#channel'],
	['#channel;', '#channel'],
	['#chan;nel', '#chan;nel'],
	['#channel.', '#channel'],
	['#chan;el;', '#chan;el'],
	['#chan[n]el,', '#chan[n]el'],
];
tests.forEach(c => {
	let match = c[0].match(/(^|\s|[@+~&%}]+)([#&][^ ,\007<>\n\r]+?)([:;.,<>\n\r]+)?$/i);
	let check = match?
		match[2] === c[1] :
		false;
	console.log(c[0], 'Passes:', check);
});
~~~

PR #567, opened 2018-09-21
Fix sidebar width when pinned in config

### LINKED ISSUE #564: Nicklist width
After latest commits nicklist width is too wide, it back to normal size clicking on nicklist button in top right menu.

userwidth

ItsOnlyBinary:
fixes #564

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -123,10 +123,6 @@ export default {
         window.addEventListener('focus', event => this.onFocus(event), false);
         window.addEventListener('blur', event => this.onBlur(event), false);
         window.addEventListener('touchstart', event => this.onTouchStart(event));
-
-        if (this.sidebarState.canPin && this.$state.setting('sidebarPinned')) {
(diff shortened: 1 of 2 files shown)


src/components/App.vue
```diff
@@ -123,10 +123,6 @@ export default {
         window.addEventListener('focus', event => this.onFocus(event), false);
         window.addEventListener('blur', event => this.onBlur(event), false);
         window.addEventListener('touchstart', event => this.onTouchStart(event));
-
-        if (this.sidebarState.canPin && this.$state.setting('sidebarPinned')) {
```
digitalcircuit: Is removing the ability to pin the sidebar from user settings/`config.json` intentional?

After a while of looking, I can't find anywhere else that [`sidebarPinned`](https://github.com/kiwiirc/kiwiirc/wiki/Configuration-Options#sidebarpinned ) is checked.  Pardon if I'm misunderstanding; this is my first time looking through.

PR #572, opened 2018-09-24
Theme - Freenode Live Event Theme

AllanOcelot:
Commiting the Freenode Live theme.

We have kept most of the theming default for this theme - as the idea is to only have a portion of Kiwi visible during the streams etc.


AllanOcelot:
Closing as no longer needed.

PR #578, opened 2018-09-26
added popout capability for the mediaviewer

srmcgann:


prawnsalad:
A slight issue where it gets messy:
1. Pop out a preview pane
2. Open the kiwi settings screen
3. Pop the preview pane back in

It shouldn't be possible to pop back in at this point as this screen has no space for it. There need to be a way to check if the media preview component can actually be opened at the point you go to pop it back in.

Issue #129, opened 2017-11-05
User mode changes not echoed back to users


KoraggKnightWolf:
When using the /mode command to alter ones own user modes, the change is not echoed back to said user. This issue appears to be general as it is confirmed after testing to at least occur on UnrealIRCd 4.0.15 and InspIRCd 2.0

KoraggKnightWolf:
This bug also applies to querying /mode #channel_name which results in the modes of a channel set to not be displayed anywhere. Confirmed to occur on InspIRCd 2.0

PR #582, opened 2018-09-28
show user mode changes in server buffer : issue #129

goqrch:
show modes change in server buffer when no channel buffer available, like for user mode ( issue : #129 )


prawnsalad:
Finally got round to understanding this PR, thanks!

PR #586, opened 2018-10-01
Experiment: Tooltips for Kiwi

AllanOcelot:
I've created this small demo to show how tooltips could possibly be used in future for Kiwi.
You can hover over the topic. settings icon and emoji options for quick examples.

These are not the final look and feel - this is merely an experiment to show how it could be done easily.


AllanOcelot:
We decided to use :title on elements instead, and let the browser decide where to place the 'tooltip'

PR #588, opened 2018-10-01
FIX: Settings and Close StateBrowser icon for mobile devices

AllanOcelot:
- Remove span from close
- Close icon in statebrowser is now absolute positioned
- Close icon same height as containerHeader height
- Settings icon clickable

### DIFF
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -1,8 +1,7 @@
 <template>
     <div class="kiwi-appsettings">
 
-        <div class="kiwi-appsettings-title" @click="closeSettings">
-            <span>{{ $t('close') }}</span>
+        <div class="kiwi-appsettings-close" @click="closeSettings">
             <i class="fa fa-times" aria-hidden="true"/>
         </div>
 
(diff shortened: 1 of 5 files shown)


ItsOnlyBinary:
I think fa-chevron-left might be better for this, as its not really closing its just collapsing away

AllanOcelot:
@ItsOnlyBinary  & @prawnsalad  

Spent some time on this today - I've redone the appSettings, sidebar and stateBrowser close to try and make everything more consistent.

There is one bug I've come across so far - if you pin the userNick list, then the spacing for the filter goes off. This could easily be solved if we add a class to the sidebar if it's pinned - then I can edit the style based on that.

( Please test using the default theme, as thats the only one the colours have been done for :) )

prawnsalad:
There is the `kiwi-container--sidebar-pinned` class added to the container once it's pinned. So you would use `.kiwi-container--sidebar-pinned .kiwi-sidebar { }`

AllanOcelot:
Awesome - I'll make a fix now :)

AllanOcelot:
@prawnsalad  @ItsOnlyBinary  - it would be good to get your feedback on this PR when you get some time :)

AllanOcelot:
This PR is now outdated, the above items have been addressed by multiple new PR's since this was opened.

PR #604, opened 2018-10-09
Message list text

AllanOcelot:
This is a new updated PR for displaying the messageList as pure text in a style similar to video streaming chats.

It's extremely useful for people who have Kiwi clients installed in smaller spaces.


prawnsalad:
One item left todo before merging: Updating themes to not style the connection messages while using the inline message layout

Issue #291, opened 2018-03-29
Colored Font makes it so that Commands are sent as channel messages beginning with a /


KoraggKnightWolf:
When utilizing a color font on KiwiIRC, it appears that the color is applied to any character, including prefixing the / with it which is probably why it is sent as a colored channel message instead of a command. 

For example using colored font to /join #channel_name sends /join #channel_name to the channel. This issue is confirmed to work with /notice as well and most likely all other commands as well.

Regards,

Koragg

PR #605, opened 2018-10-10
Fix selected colour not remaining after sending a message

### LINKED ISSUE #130: Text color temporarily resets to black when text entry field loses focus
Text color is temporarily reset after an empty text input loses focus. 

Steps to reproduce:
1. Set text color.
2. Send something. Note that the text is the color you selected.
3. Click out of the text input, change channels, or something else to lose text area focus.
4. Click back in, send something. Note that the text is black again.
5. Send another thing without losing focus on text entry. Note that text is the color you selected again.

### LINKED ISSUE #592: Text colors being reset when sending the message
When we choose a text color, after hitting Enter to send the message, the color is reset.

I'm facing this issue in Chrome for Linux and Chrome for Android. 

Cheers.

ItsOnlyBinary:
fixes #592
fixes #130

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -450,7 +450,6 @@ export default {
             this.history_pos = this.history.length;
 
             this.$refs.input.reset();
-            this.$refs.input.focus();
         },
         historyBack() {
             if (this.history_pos > 0) {
--- a/src/components/utils/IrcInput.vue
+++ b/src/components/utils/IrcInput.vue
@@ -14,6 +14,7 @@
             @mouseup="updateValueProps();"
             @click="$emit('click', $event)"
             @paste="onPaste"
+            @focus="onFocus()"
         />
     </div>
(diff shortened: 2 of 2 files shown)


ItsOnlyBinary:
Also includes fixes for....
Applying colour after removing placeholder (empty input box)
fixes #291 Colored Font makes it so that Commands are sent as channel message

utoxin:
Is there a reason this hasn't been approved yet? My users are all experiencing this, and it would be nice to have a fix.

ItsOnlyBinary:
This fix requires some more research, as there maybe a better fix

TheBeerbarian:
I like the bug that the color is reset to default after sending a message.  If it is fixed I would like the current bug to be a feature that can be a configured option.
With self hosted platforms used for my IRCd, I rather discourage people from constantly talking in color and rather it reset after a message.
Thanks. :)

PR #606, opened 2018-10-12
Fix nick & pass not updating after failed connection attempt using Welcome.vue

### LINKED ISSUE #559: attempt to enter a new nickname during a reject
hi,

It's impossible to enter a new nickname when a server rejects you a first time

http://imgur.com/43X6BEZl.png

ItsOnlyBinary:
fixes #559

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -185,39 +185,34 @@ export default {
                 return;
             }
 
-            let net;
-            if (!this.network) {
-                let netAddress = _.trim(options.server);
+            let netAddress = _.trim(options.server);
 
-                // Check if we have this network already
-                net = state.getNetworkFromAddress(netAddress);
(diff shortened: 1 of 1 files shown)


src/components/startups/Welcome.vue
```diff
@@ -185,39 +185,34 @@ export default {
...
+                port: options.port,
+                tls: options.tls,
+                password: this.password,
+                encoding: _.trim(options.encoding),
+                direct: !!options.direct,
+                path: options.direct_path || '',
+                gecos: options.gecos,
+            });
+
+            if (!this.network && options.recaptchaSiteId) {
```
ItsOnlyBinary: I added !this.network so it would only ask the first time as that's what it was originally doing, not sure if this is correct due to not having recaptcha setup.

Does recaptcha time out? would a second connection attempt work with the first recaptacha answer?

PR #612, opened 2018-10-18
Add eslint rules folder and class-name-prefix rule

ItsOnlyBinary:


prawnsalad:
This is sweet.

Can it not be loaded at runtime, ie. not via package dependancies? Adding a module dependancy within itself is very... weird and confusing.

ItsOnlyBinary:
I've not found a way to do that, but I will look into it some more on monday

PR #614, opened 2018-10-19
Fix item.id in plugin tabs conflicting with TabbedView parent

ItsOnlyBinary:
item.id was conflicting here: https://github.com/kiwiirc/kiwiirc/blob/master/src/components/utils/TabbedView.vue#L3


ItsOnlyBinary:
the other option would be to append something to :key="a" in TabbedView.vue

prawnsalad:
I'm not sure how that would cause a key conflict in TabbedView.vue as there's only ever one key in use here?

PR #627, opened 2018-10-25
Add ability for plugins to override user colours

ItsOnlyBinary:


src/helpers/TextFormatting.js
```diff
@@ -143,33 +143,38 @@ export function linkifyChannels(word) {
 
 export function linkifyUsers(word, userlist) {
     let ret = '';
-    let nick = '';
+    let user = '';
```
prawnsalad: This isn't a string, but a user object. So this should be initialised as `null`

src/helpers/TextFormatting.js
```diff
@@ -143,33 +143,38 @@ export function linkifyChannels(word) {
...
 
-    let escaped = _.escape(nick);
-    let colour = createNickColour(nick);
-    ret = `<a class="kiwi-nick" data-nick="${escaped}" style="color:${colour}">${escaped}</a>`;
+    let escaped = _.escape(user.nick);
+    let colour = user.colour;
+
+    ret = `<a class="kiwi-nick" data-nick="${escaped}"`;
+    if (colour) {
+        ret += `style="color:${colour}"`;
```
prawnsalad: This *shouldn't* be an issue but it smells a bit funny. There is no space between the HTML attributes when put together. Adding a space to the end of the string on line 173 and 175 will solve this.

src/libs/state.js
```diff
@@ -1031,6 +1032,7 @@ const state = new Vue({
                     realname: user.realname || '',
                     modes: user.modes || '',
                     away: user.away || '',
+                    colour: TextFormatting.createNickColour(user.nick),
```
prawnsalad: This going to kill performance as you could get a nicklist of thousands of nicks and colour generation is expensive. This should be generated only when needed

prawnsalad:
Any reason why the whois data are set as non-enumerables? https://github.com/kiwiirc/kiwiirc/pull/627/files#diff-2c38cabff66aa2d40fac9c089fa9ec3aR18

`import { def } from '@/libs/state/common';` can just be `import { def } from './common';` since they're strictly related

PR #633, opened 2018-10-29
Add setting nicklistSortByState

### LINKED ISSUE #434: Option to order the nicklist with away users last

ItsOnlyBinary:
closes #434

### DIFF
--- a/src/components/Nicklist.vue
+++ b/src/components/Nicklist.vue
@@ -126,6 +126,16 @@ export default {
                     modesA.length === 0 &&
                     modesB.length === 0
                 ) {
+                    // Compare away status
+                    if (this.$state.setting('nicklistSortByState')) {
+                        if (a.away && !b.away) {
+                            return 1;
+                        }
+                        if (!a.away && b.away) {
+                            return -1;
+                        }
+                    }
+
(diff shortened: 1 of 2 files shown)


ItsOnlyBinary:
prawn let me know if you like the theme additions, if so i can add them to the other themes ;)

prawnsalad:
I would have expected this to display all away nicks at the top, but it actually sorts by prefix first, and then away status. Do we know what the norm is here for other clients?

ItsOnlyBinary:
Not sure on norm for other clients but, It would not make sense to have away users at the top as we are not interested in away users only in active users. also its customary to have users with chan status' at the top

ItsOnlyBinary:
rebased

prawnsalad:
This looks OK now I'm looking at it though I'm not so sure about the `nicklistSortByState` name. Not entirely sure but maybe something like `nicklistOrderedAway`?

ItsOnlyBinary:
nicklistGroupAway  ?

prawnsalad:
That works. Also to default to `false` so we don't suddenly change everyones nicklist that they're used to.

westor7:
I personally think `nicklistSortByState` is more clear here.

PR #641, opened 2018-10-31
Improve german translations

fechnert:
Improve existing german translations and add missing ones


ItsOnlyBinary:
Unfortunately we use crowdin to manage our translations, so this cannot be merged. if possible could you please make the changes here: https://crowdin.com/project/kiwiirc

PR #642, opened 2018-11-01
Add option to hide built-in emoji picker && Auto hide emoji picker on touchscreen devices

ItsOnlyBinary:


Shillos:
I have been wanting this hidden, because I use the emojis plugin instead.

Shillos:
Can i set this in my config as showEmojiPicker: false would that work?

ItsOnlyBinary:
yes :)

src/components/ControlInput.vue
```diff
@@ -147,6 +151,9 @@ export default {
         shouldShowSendButton() {
             return this.$state.ui.is_touch || this.$state.setting('showSendButton');
         },
+        showEmojiPicker() {
```
prawnsalad: This should be renamed to `shouldShowEmojiPicker` for consistency

ItsOnlyBinary: fixed with force push ;)

PR #643, opened 2018-11-01
added auto-hide emoji picker on mobile devices

srmcgann:


prawnsalad:
@srmcgann did you notice that this commit failed?

Also, I'm not sure if this was discussed already, but is it not easier to simply check for a touch screen? Keeping up with mobile device names/agents would be a pain.

srmcgann:
This is not an ideal solution to the issue. Recommend going with PR #642 instead.

PR #644, opened 2018-11-01
Customise fontAwesome to make it smaller

ItsOnlyBinary:


ItsOnlyBinary:
This was built using https://icomoon.io/app
[fontAwesome-kiwi.zip](https://github.com/kiwiirc/kiwiirc/files/2539413/fontAwesome-kiwi.zip)
The zip includes a json file for importing the selection, it currently contains 62 icons

AllanOcelot:
@ItsOnlyBinary  - good idea, how would we go about updating it if we added in a new Icon?

ItsOnlyBinary:
You can just import the config json file from the zip into the website and add the wanted extra icons, then download the new zip and add the files to kiwi

prawnsalad:
I do like the idea for this but a lot of plugins make use of random fa icons too, so this would limit what plugins can use.

PR #649, opened 2018-11-03
Temp fix: Pinning sidebar

AllanOcelot:
User complained that unpinning the sidebar was not intuitive enough, this PR is a temp fix whilst I redesign and think about how better to implement the pinning.


prawnsalad:
Closing this in favour of the pinning toggle PR https://github.com/kiwiirc/kiwiirc/pull/650

PR #653, opened 2018-11-06
add support for newer ways to get user accounts

ItsOnlyBinary:
uses: 
-  account-notify
-  extended-join
-  WHOX

requires https://github.com/kiwiirc/irc-framework/pull/180

### DIFF
--- a/src/libs/IrcClient.js
+++ b/src/libs/IrcClient.js
@@ -413,12 +413,19 @@ function clientMiddleware(state, network) {
             }
 
             let buffer = state.getOrAddBufferByName(networkid, event.channel);
-            state.addUserToBuffer(buffer, {
+            let user = {
                 nick: event.nick,
                 username: event.ident,
                 host: event.hostname,
                 realname: event.gecos,
-            });
+            };
+
+            // Add account to the user object if extended-join is enabled
+            if (client.network.cap.isEnabled('extended-join')) {
+                user.account = event.account || '';
+            }
+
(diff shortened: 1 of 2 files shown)


src/libs/IrcClient.js
```diff
@@ -413,12 +413,19 @@ function clientMiddleware(state, network) {
...
+            let user = {
                 nick: event.nick,
                 username: event.ident,
                 host: event.hostname,
                 realname: event.gecos,
-            });
+            };
+
+            // Add account to the user object if extended-join is enabled
+            if (client.network.cap.isEnabled('extended-join')) {
```
prawnsalad: This could simply be `if (user.account) {` so that it always adds the account if it's available

src/libs/IrcClient.js
```diff
@@ -633,6 +644,7 @@ function clientMiddleware(state, network) {
                         username: user.ident || undefined,
                         away: user.away ? 'Away' : '',
                         realname: user.real_name,
+                        account: user.account === 0 ? '' : user.account,
```
prawnsalad: Why would `user.account` ever be `0`?

ItsOnlyBinary: user account is 0 when there is no account, as whox is space separated it has to return something.
i wasnt sure if this should be in irc-fw or here

prawnsalad: Then that should be checked in irc-fw and either not set the account on the event, or set it as an empty string

PR #659, opened 2018-11-07
Translate autocompleteCommands.js

ItsOnlyBinary:


src/components/ControlInput.vue
```diff
@@ -486,9 +487,12 @@ export default {
             if (opts.commands) {
                 let commandList = [];
                 autocompleteCommands.forEach((command) => {
+                    let desc = command.description.startsWith('command_') ?
+                        TextFormatting.t(command.description) :
+                        command.description;
```
ItsOnlyBinary: I've added this so the autocompleteCommands file can be overridden with static strings

prawnsalad: This should be a comment in the code itself

src/helpers/TextFormatting.js
```diff
@@ -447,7 +447,7 @@ export function t(key, options) {
     let val = i18next.t(key, options);
     if (!val) {
         let opts = options || {};
-        opts.lng = 'en-us';
+        opts.lng = 'dev';
```
prawnsalad: You would never want to show the dev language in production. It falls back to en-us as that is the default translation that we know will always be shipped complete

PR #660, opened 2018-11-08
Added file, Kiwi.vue, to be used by external Vuejs apps. Turns whole Kiwi app into vue component. Closes #647

### LINKED ISSUE #647: Embedding Kiwi IRC inside a Vue project
I'm looking for a way to embed Kiwi into another Vue project. Without customizing the config options I've been able to do so by using an <iframe> to embed but haven't been able to pass custom kiwiConfig to it. Being written in Vue itself I would think there would be an easier way to instantiate and pass the configuration but I haven't been able to figure it out.

srmcgann:


prawnsalad:
Keeping this open for now depending on the outcome of https://github.com/kiwiirc/kiwiirc/pull/664

srmcgann:
Solution is better implemented in #664. Closing.

PR #664, opened 2018-11-14
implemented optional build config for exporting kiwi as a vue component. Closes #647

### LINKED ISSUE #647: Embedding Kiwi IRC inside a Vue project

srmcgann:


package.json
```diff
@@ -14,7 +15,8 @@
     "lint:stylelint": "stylelint \"src/**/*.{vue,htm,html,css,sss,less,scss}\"",
     "lint": "npm run lint:eslint && npm run lint:stylelint",
     "lint:fix": "npm run lint:eslint -- --fix && npm run lint:stylelint -- --fix",
-    "gendocs": "rm -rf srcdocs/ && jsdoc -c jsdoc.json"
+    "gendocs": "rm -rf srcdocs/ && jsdoc -c jsdoc.json",
+    "install": "if [ $(basename $(dirname $(pwd))) = \"node_modules\" ];then node build/buildModule.js;fi"
```
prawnsalad: Does this need to be called `install`? As it's just a different build option maybe `build:module` is better suited?

The build script filename has a redundant `build` in it, it could just be `build/module.js`. Then this script can just be `build/module.js`.

build/webpack.base.conf.js
```diff
@@ -25,13 +25,13 @@ module.exports = {
   entry: {
     app: [
         'core-js/fn/promise', // required by the webpack runtime for async import(). babel polyfills don't help us here. ie11
-        './src/main.js'
+        './src/main.' + (process.env.NODE_ENV === 'module' ? 'vue' : 'js')
```
prawnsalad: `(process.env.NODE_ENV === 'module' ? 'vue' : 'js')` could do with being moved to the top under `const outputExt = (process.env.NODE_ENV === 'module' ? 'vue' : 'js');`  to make this config section clearer to read

src/main.vue
```diff
@@ -0,0 +1,418 @@
...
+
+            // Add any properties to the raw element
+            if (binding.value.props) {
+                Object.keys(binding.value.props).forEach((key) => {
+                    rawEl[key] = binding.value.props[key];
+                });
+            }
+        }
+    },
+});
```
prawnsalad: Do these vue mixins and directives get added to the parent application? This could cause conflicts

src/main.vue
```diff
@@ -0,0 +1,418 @@
...
+import GlobalApi from '@/libs/GlobalApi';
+import { AudioManager } from '@/libs/AudioManager';
+import { SoundBleep } from '@/libs/SoundBleep';
+
+// Global utilities
+import '@/components/utils/TabbedView';
+import '@/components/utils/InputText';
+import '@/components/utils/IrcInput';
+import '@/components/utils/InputPrompt';
+import '@/components/utils/InputConfirm';
```
prawnsalad: Do these global components get added to the parent application? This could cause conflicts

src/main.vue
```diff
@@ -0,0 +1,418 @@
...
+        el: '#kiwi-app',
+        render: h => h(StartupError),
+    });
+}
+export default{
+    props: ['kiwiConfig'],
+    mounted() {
+        window.kiwiConfig = () => this.kiwiConfig;
+    },
+    template: '<div id="kiwi-app"></div>',
```
prawnsalad: main.vue exports a component with a hardcoded `template`, but it also contains a `<template>` block? Which is in use?

src/main.vue
```diff
@@ -0,0 +1,418 @@
...
+    /* eslint-disable no-new */
+    new Vue({
+        el: '#kiwi-app',
+        render: h => h(StartupError),
+    });
+}
+export default{
+    props: ['kiwiConfig'],
+    mounted() {
+        window.kiwiConfig = () => this.kiwiConfig;
```
prawnsalad: It looks like it's reading the config from the components `kiwiConfig` prop, but then it's setting that to `window.kiwiConfig()` and that's where it's read from in the kiwi app? Why not just use the property directly?

src/main.vue
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /* eslint-disable no-new */
+    new Vue({
+        el: '#kiwi-app',
+        render: h => h(StartupError),
+    });
+}
+export default{
+    props: ['kiwiConfig'],
```
prawnsalad: This component is specifically a kiwi component so the `kiwi` part here is redundant. It should be be named simply `config`.

src/main.vue
```diff
@@ -0,0 +1,418 @@
+<template>
+    <div id="kiwi-component-container">
+        <div id="kiwi-app"/>
+    </div>
+</template>
+<script>
```
prawnsalad: Almost all of main.vue is duplicated from main.js which is going to cause a fair bit of maintenance. Can we re-use something here?

With that, some parts of this code is not needed such as the loggers `window.href.location` checking

package.json
srmcgann: "install" is an npm hook that triggers the build when the module is installed into the parent app. One of these is needed here:
https://docs.npmjs.com/misc/scripts

As per the second part of this request, I've renamed the build script accordingly.

build/webpack.base.conf.js
srmcgann: done. (except i used "inputExt" instead of "outputExt" since this file is a source file, not an output file.

src/main.vue
srmcgann: Oops. Yes, there was a duplicate template definition. I've removed one.

src/main.vue
srmcgann: done.

src/main.vue
srmcgann: I think they do, but there's not really a way to prevent that. Perhaps a warning in the docs would suffice?

src/main.vue
srmcgann: I think they do, but there's not really a way to prevent that. Perhaps a warning in the docs would suffice?

src/main.vue
srmcgann: This is done.

prawnsalad:
Unfortunately closing this PR so that we keep the app accessible via an iframe only for embedding. The added complexity of being a component isn't worth the amount this would actually be used.

PR #672, opened 2018-11-20
Make network staff and channel ops exempt from block new pm's

### LINKED ISSUE #285: "Block new private messages" should still allow network staff messages
Network staff and any user on a shared channel with mode +h or above.

ItsOnlyBinary:
Closes #285

### DIFF
--- a/src/helpers/Misc.js
+++ b/src/helpers/Misc.js
@@ -4,7 +4,6 @@
 
 import _ from 'lodash';
 import strftime from 'strftime';
-import state from '@/libs/state';
 
 /**
  * Extract an array of buffers from a string, parsing multiple buffer names and channel keys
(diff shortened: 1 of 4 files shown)


src/helpers/Misc.js
```diff
@@ -207,3 +208,25 @@ export function dateIso(date) {
...
+        let buffer = buffers[i];
+        if (buffer.isUserAnOp(nick)) {
+            return true;
+        }
+    }
+
+    let user = state.getUser(networkid, nick);
+    // if we have not seen or whois the user they might be a network oper
+    if (!user || !user.hasWhois) {
+        return null;
```
prawnsalad: Any reason this is returning `null` while everywhere else is a boolean?

prawnsalad:
I'm getting the feeling that this may be tidier on the NetworkState class? And called something like `isNewPmBlockedFromNick(nick)`?

PR #690, opened 2018-12-01
Fix #689 - The display problem of the servers setting page

### LINKED ISSUE #689: The changes of the InputText component cause a display problem
The changes of the InputText component to fit the welcome pages cause a display problem of the old servers setting pages where the InputText component used before.

I'm dealing with the problem and will PR later.

remini1998:
Fix the display problem of the servers setting page.

### DIFF
--- a/src/components/NetworkSettings.vue
+++ b/src/components/NetworkSettings.vue
@@ -291,11 +291,15 @@ export default {
     font-size: 0.8em;
 }
 
+.kiwi-networksettings .u-input-text input {
+    height: 40px;
+}
+
 .kiwi-networksettings input[type='text'],
     .kiwi-networksettings input[type='password'],
     .kiwi-networksettings input[type='email'],
     .kiwi-networksettings textarea,
-    .kiwi-networksettings .input-text input {
+    .kiwi-networksettings .u-input-text input {
     clear: both;
     width: 100%;
     height: 40px;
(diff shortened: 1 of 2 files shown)


ItsOnlyBinary:
Thank you for your contribution, but this will need some changes before it can be merged.

There is no need to add any css, all the needed ccs is already there, but some are still missing the added `u-`

it should be fixable by just editing the css in NetworkSettings.vue, no need to touch other files

remini1998:
@ItsOnlyBinary 
Yep, I think u r right. It's my mistake. Plz wait for a moment~

ItsOnlyBinary:
i think its around 6 `u-` that need adding (including the one you already caught ;)

remini1998:
@ItsOnlyBinary 
Thanks for pointing out! :D
I notice another problem while fixing.
It's about the "line-height: 40px".
This line make Safari cursor not look pretty (because the cursor has a 40px-height which is more longer than the character), and it do nothing on Chrome and Firefox.
Is it a feature or something to make it perform better on IE or Edge?
If not, I'd like to remove that line to make it look better on Safari. Would I?

ItsOnlyBinary:
I would recommend doing that as a separate pull-request, explain why the change in the comments and maybe attach a before and after screenshot to show how it effects safari. Unfortunately i don't have safari to test it and not sure what effects it could have on ff, chrome or ie (I could test those though)

remini1998:
@ItsOnlyBinary 
I think it's done.
Thanks for review.
I'm going to deal with the next problem

PR #702, opened 2018-12-07
Allow sidebar in queries

### LINKED ISSUE #685: show user sidebar in private messages
hi,

Can you display the sidebar of the user in private messages, it is more pleasant and intuitive for the user to have it in private messages, for example: http://imgur.com/GgNKEF8l.png

ItsOnlyBinary:
This changes the api slightly

closes #685

### DIFF
--- a/src/components/SidebarState.vue
+++ b/src/components/SidebarState.vue
@@ -6,11 +6,20 @@ import Vue from 'vue';
 export default Vue.extend({
     data() {
         return {
-            sidebarOpen: false,
-            // sidebarSection may be either '', 'user', 'settings', 'nicklist', 'about'
-            sidebarSection: '',
-            sidebarUser: null,
             activeComponent: null,
+            // sidebarSection may be either '', 'user', 'settings', 'nicklist', 'about'
+            channelState: {
+                allowedSections: ['user', 'settings', 'nicklist', 'about'],
+                sidebarOpen: false,
+                sidebarUser: null,
+                sidebarSection: '',
+            },
+            queryState: {
+                allowedSections: ['user'],
+                sidebarOpen: null,
+                sidebarUser: null,
+                sidebarSection: 'user',
+            },
         };
     },
     computed: {
@@ -23,6 +32,43 @@ export default Vue.extend({
         isOpen() {
             return this.sidebarOpen && this.$state.ui.app_width > 769;
         },
+        sidebarState() {
+            let buffer = this.$state.getActiveBuffer();
+            return buffer.isChannel() ? this.channelState : this.queryState;
+        },
+        sidebarOpen: {
+            get() {
+                let state = this.sidebarState;
+                return state.sidebarOpen;
+            },
+            set(val) {
+                let state = this.sidebarState;
+                state.sidebarOpen = val;
+            },
+        },
+        sidebarUser: {
+            get() {
+                let state = this.sidebarState;
+                return state.sidebarUser;
+            },
+            set(val) {
+                let state = this.sidebarState;
+                state.sidebarUser = val;
+            },
+        },
+        sidebarSection: {
+            get() {
+                let state = this.sidebarState;
+                if (!state.sidebarOpen) {
+                    return '';
+                }
+                return state.sidebarSection;
+            },
+            set(val) {
+                let state = this.sidebarState;
+                state.sidebarSection = val;
+            },
+        },
     },
     created() {
         this.listen(this.$state, 'sidebar.component', (component) => {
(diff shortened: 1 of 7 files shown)


src/components/SidebarState.vue
```diff
@@ -6,11 +6,20 @@ import Vue from 'vue';
...
+            // sidebarSection may be either '', 'user', 'settings', 'nicklist', 'about'
+            channelState: {
+                allowedSections: ['user', 'settings', 'nicklist', 'about'],
+                sidebarOpen: false,
+                sidebarUser: null,
+                sidebarSection: '',
+            },
+            queryState: {
+                allowedSections: ['user'],
+                sidebarOpen: null,
```
prawnsalad: Why is this null instead of a bool? I see some other places that compares this with null too for some reason

src/components/SidebarState.vue
```diff
@@ -34,72 +80,125 @@ export default Vue.extend({
...
+                this.channelState.sidebarOpen = true;
+                this.queryState.sidebarOpen = true;
             }
         });
     },
     methods: {
         section() {
+            if (this.sidebarOpen === null) {
+                this.sidebarOpen = this.channelState.sidebarOpen;
+            }
```
prawnsalad: I can't tell what this if block is doing? If null, then it... calls the getter to get sidebarOpen and then sets itself back to its own value?

src/components/SidebarState.vue
```diff
@@ -34,72 +80,125 @@ export default Vue.extend({
...
+                ) {
+                    let user = this.$state.getUser(buffer.getNetwork().id, buffer.name);
+                    // if the query sidebar is open and not the current user, switch users
+                    this.showSidebar('user', { user: user });
+                }
+                return this.sidebarSection;
             }
 
             return '';
         },
```
prawnsalad: This `section` method now has introduced side effects. It should simply return what section shoudl be displayed in the sidebar if it is open, however you now have it opening the sidebar in some cases

src/components/SidebarState.vue
```diff
@@ -34,72 +80,125 @@ export default Vue.extend({
...
-        showUser(user) {
+        showSidebar(section, opts) {
             this.activeComponent = null;
-            this.sidebarUser = user;
-            this.sidebarOpen = true;
-            this.sidebarSection = 'user';
+
+            // Only set sidebar section if its in the allowed list
+            if (this.sidebarState.allowedSections.includes(section)) {
```
prawnsalad: We use the return early pattern to prevent huge nested blocks and improves readability. So this line would be:
~~~javascript
if (!this.sidebarState.allowedSections.includes(section)) {
    return;
}
~~~

src/components/SidebarState.vue
```diff
@@ -34,72 +80,125 @@ export default Vue.extend({
...
 
             return '';
         },
         close() {
             this.sidebarOpen = false;
-            this.sidebarSection = '';
             this.sidebarUser = null;
         },
-        showUser(user) {
+        showSidebar(section, opts) {
```
prawnsalad: `showSidebar()` has now replaced `showUser()` while all the other `show*()` methods still exist, making this inconsistent. Either the API should use `showSidebar()` or `showUser()` should be be re-implemented. Re-implementing it would be easier for devs further down the line.

src/components/SidebarState.vue
```diff
@@ -34,72 +80,125 @@ export default Vue.extend({
...
-                this.showBufferSettings();
+            this.toggleSidebar('settings');
         },
         toggleAbout() {
-            this.sidebarSection === 'about' ?
-                this.close() :
-                this.showAbout();
+            this.toggleSidebar('about');
+        },
+        getUserFromOpts(opts) {
```
prawnsalad: What's the point of this function? It reads the `opts.user` and then looks it up again from state? We pass the user object already to sidebar functions.

src/components/SidebarState.vue
ItsOnlyBinary: This will cause the sidebar to open in queries if it was open in channel when the first query is opened

src/components/SidebarState.vue
ItsOnlyBinary: related to above, null means query has not been opened yet, so get the state from channel

src/components/SidebarState.vue
ItsOnlyBinary: This ensures that the sidebar shows the user for the current query. ie switching queries will always show the correct user

ItsOnlyBinary:
Closed in favour of #781

PR #705, opened 2018-12-07
skip default input handler if event.handled was set by another handler

vith:


prawnsalad:
`handleMessage()` handles a message type command, privmsg/action/notice, https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/InputHandler.js#L177

The handled check is here https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/InputHandler.js#L107 which is directly after it has called any plugin handlers.

vith:
What I'm trying to enable here is for a plugin to get a first look at non-command input and have the option of preventing the default /msg handler from running.

I've looked over it again but I still don't see any existing way to do this. If the typed input is not a command, it is rewritten to be /msg in channels/etc. here:  https://github.com/kiwiirc/kiwiirc/blob/6dd459853529cc9e60385e7653c0b676ff7388ff/src/libs/InputHandler.js#L59-L71

So eventually the `/msg` handler will be called through this emit: https://github.com/kiwiirc/kiwiirc/blob/6dd459853529cc9e60385e7653c0b676ff7388ff/src/libs/InputHandler.js#L104-L105

The event emitter we're using doesn't have any built in way for a handler to make it break out of the loop over the registered listeners.

vith:
If overloading the `.handled` property to accomplish this isn't desired, would a new `'pre-input'` event be better? With a `cancel` function passed to the handlers as an extra argument that would disable calling the `'input.command.*'` handlers if called.

PR #707, opened 2018-12-08
Be able to custom the nickname check rules of the welcome page

remini1998:
There are some irc-servers can totally support utf-8 such as Oragono.
To connect the irc-server like that, English only username is unnecessary, so I think it's a good idea to add the feature of custom the nickname check rules of the welcome page.
That's what this P.R. is.


prawnsalad:
Oragono supports UTF8 in nicks? How does it handle the abuse issues of nicks masquerading as others?

remini1998:
> Oragono supports UTF8 in nicks? How does it handle the abuse issues of nicks masquerading as others?

I don't know either... But I have tried and it does.

prawnsalad:
I've changed this some more to make the config easier. `startupOptions.nick_format` can now be a full regex pattern such as `'/@domain.com$/i'` or a simple string such as `'kiwi'` which turns into `/kiwi/i`.

PR #709, opened 2018-12-11
Fix loading html plugins on IE11

### LINKED ISSUE #708: html plugins break kiwi on IE11
the plugin loader for html files uses fetch() which does not exist in IE11

i have tried polyfilling with https://github.com/github/fetch

and replacing the following forEach with a for (x of y)
https://github.com/kiwiirc/kiwiirc/blob/master/src/main.js#L265

this is getting me:
SCRIPT1003: Expected ':'
https://gist.github.com/ItsOnlyBinary/1754abd319ea4c9233d8db083760a79d

ItsOnlyBinary:
closes #708

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "webpack-bundle-analyzer": "^2.9.0",
     "webpack-dev-server": "^2.11.1",
     "webpack-merge": "^0.14.1",
+    "whatwg-fetch": "^3.0.0",
     "xhr": "^2.3.2"
   },
   "devDependencies": {
--- a/src/main.js
+++ b/src/main.js
@@ -4,6 +4,9 @@ import i18next from 'i18next';
 import i18nextXHR from 'i18next-xhr-backend';
 import VueI18Next from '@panter/vue-i18next';
 
+// fetch polyfill
+import 'whatwg-fetch';
+
 import AvailableLocales from '@/res/locales/available.json';
 import FallbackLocale from '@/../static/locales/en-us.json';
 import App from '@/components/App';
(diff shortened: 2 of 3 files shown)


prawnsalad:
It might be cleaner to convert the NodeList to an array instead so that you don't to fight against the linter. `let scripts = Array.apply(null, el.querySelectorAll('script'))` should handle that.

I think @vith had actually resolved the fetch() issue from within webpack already, thoughts @vith?

ItsOnlyBinary:
```
  ✘  http://eslint.org/docs/rules/prefer-spread         Use the spread operator instead of '.apply()'

  C:\Users\Neil\Code\kiwiirc\kiwiirc\src\main.js:268:35
                      let scripts = Array.apply(null, el.querySelectorAll('script'));
                                     ^

  ✘  http://eslint.org/docs/rules/no-restricted-syntax  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations

  C:\Users\Neil\Code\kiwiirc\kiwiirc\src\main.js:271:21
                      for (let limitedScr of scripts) {
                       ^
```

vith:
https://github.com/kiwiirc/kiwiirc/pull/624 and https://github.com/kiwiirc/kiwiirc/pull/629 are the relevant PRs of mine.

In the first one I was going to polyfill fetch but after enabling babel-polyfill in the second attempt it was working without the fetch polyfill. Not sure what's different in this case, but if something's calling fetch it's reasonable to polyfill it.

I was injecting it as a webpack entrypoint rather than importing it into one of our source files. I think that may be necessary to enable the clientside webpack code to load async chunks. But we aren't using that feature so we can revisit that later if needed.

PR #717, opened 2018-12-16
FIX: Issue 712 - Change ban/kick user actions colour on dark theme.

AllanOcelot:
- Remove the background colour for the OP user actions, replace with border colour instead, to increase legibility


ItsOnlyBinary:
dont forget nightswatch ;)

AllanOcelot:
@ItsOnlyBinary  - just done :)

PR #728, opened 2018-12-18
Add message bodyTemplate

ItsOnlyBinary:
Allow adding a component to the message body


prawnsalad:
Just a html comment above the following saying that the `message.template` component is checked first for a custom component, and then each message layout checks for a `message.bodyTemplate` custom component to apply only to the body area after. https://github.com/kiwiirc/kiwiirc/blob/master/src/components/MessageList.vue#L32

PR #729, opened 2018-12-18
Make message. event consistent

ItsOnlyBinary:
message.new does not match other message. events

it directly sends the message where as others send events with the message inside


prawnsalad:
Closing in favour of https://github.com/kiwiirc/kiwiirc/commit/662d2a36e874df46efd72dae612a32e01bd095a2

Issue #733, opened 2018-12-21
Issues on latest build


Shillos:
Here is my gist https://gist.github.com/Shillos/b7fc3788f58885b88ff734cad36e63b0

AllanOcelot:
Hi @Shillos  !

Thanks for spotting this, it seems someone recently merged an older PR into Kiwi, one that was made before the class-name linting checks were put in place. PR #735  fixes this, and should be merged in when @prawnsalad comes online.

AllanOcelot:
Pull request fixed this issue, so closing it. Thank you for reporting it :)  👍

PR #735, opened 2018-12-21
FIX: issue #733 

AllanOcelot:
- Prefixed classnames with kiwi to fix linting error


prawnsalad:
Remember that classnames should be int he form `kiwi-[component]-[thing]--[modifier]`. So in this case, `kiwi-messageinfo-kick-user`.

Then also this needs updating in the theme files

PR #739, opened 2018-12-22
Don't switch focus to text input when shift key is used

### LINKED ISSUE #737: Pressing shift causes the input box to focus
Some users use shift and arrow keys to adjust a text selection, it would be ideal if this did not cause the input box to gain focus

ItsOnlyBinary:
fixes #737

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -181,6 +181,18 @@ export default {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


prawnsalad:
This now breaks shift+letter for a capital letter

ItsOnlyBinary:
Why did i not see that, facepalm

PR #741, opened 2018-12-22
Fix autocomplete replacing words when no matches found

ItsOnlyBinary:
I think this is a better solution than #740 because it stop the borders of the autocomplete box loading

I kept #740 open though because im not sure if the code in that area needs cleaning up

### DIFF
--- a/src/components/AutoComplete.vue
+++ b/src/components/AutoComplete.vue
@@ -130,6 +130,10 @@ export default {
         },
     },
     mounted: function mounted() {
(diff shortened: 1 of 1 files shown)


prawnsalad:
If you start auto completing and there's no results, you can always hit backspace and the auto completing list re-populates. This PR cancels the auto complete so that this is no longer possible.

ItsOnlyBinary:
I've not managed to reproduce this issue in master so closing this pr and #740

PR #747, opened 2018-12-30
FIX: Build lint error + Mobile statebrowser close icon display

AllanOcelot:
This PR fixes the position of the 'close' icon in the statebrowser. Makes the styling between the KiwiSettings and Close icon consistent too.

I've included a fix for the messageInfo linting error - as I didn't want this build to fail. 

Two birds, one stone - as they say :)


prawnsalad:
What was the reasoning behind this close button? As the server buffer can't be closed.

Then the linting fixes are in https://github.com/kiwiirc/kiwiirc/pull/735 so is this supposed to supersede that PR?

PR #750, opened 2019-01-02
Update: Messagecount styled to act as open/close button for statebrowser

AllanOcelot:
- Reposition the messagecount
- Apply messagecount new styling to all theme files
- remove close button from statebrowser, as it is no longer needed.

### DIFF
--- a/src/components/StateBrowserNetwork.vue
+++ b/src/components/StateBrowserNetwork.vue
@@ -287,6 +287,10 @@ export default {
             network.showServerBuffer('channels');
         },
         showBufferSettings(buffer) {
+            if ( window.innerWidth < 770 ) {
+                state.$emit('statebrowser.hide');
+                console.log(window.innerWidth);
+            }
             this.setActiveBuffer(buffer);
             this.sidebarState.showBufferSettings();
         },
--- a/src/components/StateBrowserNetwork.vue
+++ b/src/components/StateBrowserNetwork.vue
@@ -287,7 +287,7 @@ export default {
             network.showServerBuffer('channels');
         },
         showBufferSettings(buffer) {
-            if (window.innerWidth < 770) {
+            if (this.$state.ui.is_narrow) {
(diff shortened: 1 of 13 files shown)


AllanOcelot:
@prawnsalad  // @ItsOnlyBinary  // @srmcgann  

I'm trying to replace the messageCount with a font-awesome icon once the stateBrowser is open, Tried to figure it out myself and the only thing I can find that looks promising is 'stateBrowserDrawOpen' ( Found within the app.vue ).

I've tried doing 'state.stateBrowserDrawOpen' in an if statement, but sadly it's not working - any tips? Am I doing this the right way?

prawnsalad:
Have you tried `sidebarState.isOpen` as used on line 5 of Container.vue?

AllanOcelot:
@prawnsalad  - Aye, tried it first, always returns false

prawnsalad:
Can you paste specifically what you've tried? That function is specifically what opens the statebrowser drawer.

https://github.com/kiwiirc/kiwiirc/blob/9d189b0f3cc8d34283e58ca1a8c0914467c7b261/src/components/Container.vue#L5

prawnsalad:
Ah re-reading your messages, you're looking for when the statebrowser is open, not the sidebar. `.kiwi-wrap--statebrowser-drawopen {}` is the CSS selector for that

src/components/StateBrowserNetwork.vue
```diff
@@ -287,6 +287,9 @@ export default {
             network.showServerBuffer('channels');
         },
         showBufferSettings(buffer) {
+            if (window.innerWidth < 770) {
```
ItsOnlyBinary: this needs to be `this.$state.ui.app_width <= 769`

prawnsalad: As of 2 minutes ago, this should now be `if (this.$state.ui.is_narrow) {`

PR #755, opened 2019-01-03
Make channel active when joining from channel list on small screens

### LINKED ISSUE #719: Channel List Improvement
48370985_1231934476947094_6593283705739935744_n

Return to Channel from the Channel List is not not very intuitive on mobiles 

[ 0 ] does not indicate to user as a return button

ItsOnlyBinary:
closes #719

### DIFF
--- a/src/components/ChannelList.vue
+++ b/src/components/ChannelList.vue
@@ -153,6 +153,13 @@ export default {
             return content.html;
         },
         joinChannel(channelName) {
+            // Switch buffer if its already exists
+            let buffer = this.$state.getBufferByName(this.network.id, channelName);
+            if (buffer) {
+                this.$state.setActiveBuffer(this.network.id, channelName);
+                return;
+            }
+
             this.$state.addBuffer(this.network.id, channelName);
             this.network.ircClient.join(channelName);
             if (this.$state.ui.app_width <= 769) {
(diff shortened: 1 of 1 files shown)


src/components/ChannelList.vue
```diff
@@ -155,6 +155,9 @@ export default {
         joinChannel(channelName) {
             this.$state.addBuffer(this.network.id, channelName);
             this.network.ircClient.join(channelName);
+            if (this.$state.ui.app_width <= 769) {
```
ItsOnlyBinary: I've used this in more than one place, not sure if it should be made into a method in state or somewhere more appropriate

ItsOnlyBinary:
This will now also switch the active buffer when join is clicked on an already joined channel

finalhit:
@ItsOnlyBinary
> on large screens it will open the channel on first click and switch on second click (this is to allow joining multiple channels without it switching away from the channel list)

Second click does not switch to the new channel, just tried.

Anyway this is very strange implementation. I am not aware of IRC clients opening channels in the background. 

I think it should always make channel active when joining from channel list, not on small screens only.

ItsOnlyBinary:
test it [here](https://builds.kiwiirc.com/kiwiirc/pull-755/) as this is a pull request so has not been merged with master

finalhit:
I tested and it looks pretty unintuitive.

It even encourages spammers. Someone could easily join hundreds of channels in no time.

PR #763, opened 2019-01-08
Scroll to start of unread messages when switching buffers

ItsOnlyBinary:


ItsOnlyBinary:
rebased

src/components/MessageList.vue
```diff
@@ -423,6 +425,38 @@ export default {
         scrollToBottom() {
             this.$el.scrollTop = this.$el.scrollHeight;
         },
+        scrollToUnread() {
+            let messageId = this.buffer.last_read ? -1 : this.filteredMessages[0].id;
+
+            // Find the message id for unread marker
+            for (let i = 0; i < this.filteredMessages.length; i++) {
+                if (this.shouldShowUnreadMarker(i)) {
+                    let idx = (i - 1 >= 0) ? i - 1 : 0;
```
prawnsalad: Could do with a quick comment on what this calculation is for

src/components/MessageList.vue
```diff
@@ -423,6 +425,38 @@ export default {
...
+                    messageId = this.filteredMessages[idx].id;
+                    break;
+                }
+            }
+
+            // No unread messages found scroll to bottom
+            let forceScroll = this.buffer.setting('force_scroll_to_bottom');
```
prawnsalad: In what case would it ever not want to go to the bottom here? scrollToUnread() must scroll to the unread messages. If there are none available then unread starts at the very bottom

src/components/MessageList.vue
ItsOnlyBinary: wish i could remember what its for :(

src/components/MessageList.vue
ItsOnlyBinary: This for for people who would like to disable the scroll to unread feature  (scroll to unread drives me mad on some channels on discord where i only want to see most recent messages)

prawnsalad: It sounds like you're mixing up disabling of this feature and skipping to the bottom when there are no unread messages available?

Issue #758, opened 2019-01-07
[Suggestion] Make Whois URL & Channels Clicable!


Shillos:
I thought this would be a good idea to make these clicable when someone does a whois to show up the channels as clickable to join the channel(s) and the url's that show up to open as a new window (so they don't leave the chat)! 
![kiwi-whois](https://user-images.githubusercontent.com/491386/50783651-312da900-12b4-11e9-9a2d-ce1379ac931f.png)

Shillos:
A quick question is there a way to change the Real Name to whatever we want reflecting our own network name?!

ItsOnlyBinary:
realname is controlled by webircgateway, so if your hosting your own yes, if your using kiwiirc.com no

Shillos:
> 
> 
> realname is controlled by webircgateway, so if your hosting your own yes, if your using kiwiirc.com no

how do i change that cos i'm hosting on my own hosting!

ItsOnlyBinary:
https://github.com/kiwiirc/webircgateway/blob/master/config.conf.example#L27

Shillos:
Thanks! I hope also to see clicable urls and channels soon on the whois!

Shillos:
Thought to close this one, while this was resolved!

PR #765, opened 2019-01-09
Linkify channels in UserBox.vue

ItsOnlyBinary:
partial #758

I'm not sure if linkifying url's in realname is a good idea

### DIFF
--- a/src/components/UserBox.vue
+++ b/src/components/UserBox.vue
@@ -72,9 +72,12 @@
                 <span v-if="user.secure" class="kiwi-userbox-whois-line">
                     {{ $t('user_secure') }}
                 </span>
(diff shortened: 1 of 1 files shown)


Shillos:
Well done for the linkification, but there is another problem (bug). When you whois someone in the userbox and the person is an op or owner clicking on those links brings you up not in a channel but a private chat that does't exist. Here some screen shots! 
kiwi-chan-link-error-1
Clicking the link with the symbols brings you in a private chat that doesn't exist.
kiwi-chan-link-error-2

Hope this will help!

ItsOnlyBinary:
thanks Shillos, could have sworn i had tested it with ops and it was only linking the channel part, i will look into it

PR #781, opened 2019-01-14
Allow sidebar in queries (single state)

### LINKED ISSUE #685: show user sidebar in private messages

### LINKED ISSUE #818: [BUG] UserBox  stays open when joining new channel
**To replicate:**
- Go to channel
- Open Nicklist
- Click on a user in Nicklist
- From Statebrowser, join new or existing channel

**Expected Behaviour**
The UserBox should close when joining / leaving channels, as the user may not be in the new room / channel you are joining.

ItsOnlyBinary:
closes #685
closes #818

### DIFF
--- a/src/components/SidebarState.vue
+++ b/src/components/SidebarState.vue
@@ -15,13 +15,13 @@ export default Vue.extend({
     },
     computed: {
         isDrawn() {
-            return this.sidebarOpen && this.$state.ui.app_width <= 769;
+            return this.sidebarOpen && this.section() && this.$state.ui.app_width <= 769;
         },
         isClosed() {
-            return !this.sidebarOpen;
+            return !this.section();
         },
         isOpen() {
-            return this.sidebarOpen && this.$state.ui.app_width > 769;
+            return this.sidebarOpen && this.section() && this.$state.ui.app_width > 769;
         },
     },
     created() {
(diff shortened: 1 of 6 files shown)


ItsOnlyBinary:
Squashed and rebased

ItsOnlyBinary:
rebased

CoryChaplin:
Is it wanted to remove sidebar.toggle ?

ItsOnlyBinary:
showSidebar() is unused from what i can tell. If thats what your referring to

src/components/SidebarState.vue
```diff
@@ -15,13 +15,13 @@ export default Vue.extend({
     },
     computed: {
         isDrawn() {
-            return this.sidebarOpen && this.$state.ui.app_width <= 769;
+            return this.sidebarOpen && this.section() && this.$state.ui.app_width <= 769;
         },
         isClosed() {
-            return !this.sidebarOpen;
+            return !this.section();
```
prawnsalad, line 21: Everywhere else checks for `this.sidebarOpen`, shouldn't this do the same?

src/components/SidebarState.vue
```diff
@@ -43,19 +43,37 @@ export default Vue.extend({
...
 
-            if (section === 'settings' && isChannel) {
-                return 'settings';
-            } else if (section === 'user' && this.sidebarUser && isChannel) {
-                return 'user';
-            } else if (section === 'nicklist' && isChannel) {
+            if (buffer.isQuery()) {
+                let user = this.$state.getUser(buffer.getNetwork().id, buffer.name);
+                if (user) {
+                    this.sidebarUser = user;
```
prawnsalad, line 56: `section()` is used for getting the current section, yet this has a side effect of setting the user. This would be better in a computed property that the template reads from instead

src/components/SidebarState.vue
```diff
@@ -43,19 +43,37 @@ export default Vue.extend({
...
+                }
+                return '';
+            }
+
+            // The following code is for channels only
+            if (!buffer.isChannel()) {
+                return '';
+            }
+
+            if (section === 'user' && this.sidebarUser) {
```
prawnsalad, line 67: This is checking for a user despite the user section handling this a few lines up. Should this be needed here?

prawnsalad:
In sidebarstate.js line 28
~~~javascript
        this.listen(this.$state, 'sidebar.component', (component) => {
            this.activeComponent = component;
            this.sidebarOpen = !!component;
            this.sidebarSection = '';
        });
~~~

If `component` is truthy it could do with setting `this.sidebarSection = 'component'` so that it will stay open. Looking at the rest of the file it looks like that's all that's needed... I think

src/components/SidebarState.vue
ItsOnlyBinary: seems this computed was no longer used

src/components/SidebarState.vue
ItsOnlyBinary: I've made better comments for why this side effect exists.

Basically its so the sidebar will show a user when switching to pm without changing the state from nicklist so when you return to a channel with a nicklist it will be shown

src/components/SidebarState.vue
ItsOnlyBinary: this extra check is for when switching from a pm back to a channel where that user is not present, it will instead show the nicklist, ive improved this to also change the state so it doesnt jump again when switching to a channel with the last user you opened.

also improved the comment here explaining why

prawnsalad:
I think the logic in SidebarState.js has become a little unwieldy with the changes it got over time which makes extending it a little awkward. It works as it is which is fine but I think we should draw out the logic for what it should actually do and how it decides what section to show and essentially re-write this files logic. The logic drawn out will also be documenting it for future as it's difficult to follow currently.

Just waiting on confirmation that this works with plugins from elephantman before merging.

CoryChaplin:
@prawnsalad tested the latest version of this and it works well including with a sidebar inside queries coming from a plugin.
👍 from me and great work @ItsOnlyBinary !

PR #785, opened 2019-01-18
Add real name (gecos) to advanced network settings.

TheBeerbarian:
Please review and advise if need any changes/additions.
Thx.


prawnsalad:
This looks good to me. Maybe moving it above the raw option as it looks a bit out of place right at the bottom?

Issue #716, opened 2018-12-14
Mobile Display Error


MrGrays:
![display error](https://user-images.githubusercontent.com/43826343/50025744-737a0800-ffde-11e8-8641-ee6f4cc6a49c.png)

Issue where on Samsung internet using a Samsung Galaxy S5 the channel name is not displayed correctly.

AllanOcelot:
I've discussed this with @ItsOnlyBinary - you are using his plugin.

I'm going to look at redeveloping the way the topic is handled, moving it over to flex layout, so it will have priority on spacing. For now I've recommended to @ItsOnlyBinary  to swap over to an icon, instead of text.

PR #788, opened 2019-01-22
Don't wrap channel names

ItsOnlyBinary:
fixes: #716


AllanOcelot:
Nice, this is exactly the fix I would have implemented 👍

ItsOnlyBinary:
still two issues with this pr: 
Hovering to reveal the full name pushes the buttons off the screen
and: https://imgur.com/dGDhQLm

prawnsalad:
Closing in favour of #796

Issue #720, opened 2018-12-16
Reconnect after disconnect button wrap


MrGrays:
![48371939_1151027181732976_1288447672918736896_n](https://user-images.githubusercontent.com/43826343/50057571-5c652280-0164-11e9-80d4-9184e43f24e4.jpg)

Reconnect to server wrap is not displaying correctly on mobile devices

prawnsalad:
Quick note from discussion with @AllanOcelot; Moving the connect button to the top right, and making it available on all buffers

AllanOcelot:
Fixed in #791

PR #791, opened 2019-01-23
FIX: #720 - Connection button in header now displays on right side

AllanOcelot:
- Made connection button appear in the top right corner of the containerHeader
- Added connection button to channels


ItsOnlyBinary:
This is no longer needed as we have NotConnected.vue

Issue #790, opened 2019-01-23
Remove the single user icon from the header


prawnsalad:
The highlight can simply stay on the nicklist icon perhaps, but a sudden new icon is confusing

![image](https://user-images.githubusercontent.com/725880/51616810-ce2e4a00-1f22-11e9-9788-dede77d07f77.png)

AllanOcelot:
Fixed in #792

PR #792, opened 2019-01-23
FIX: #790 -  Remove single user icon from containerHeader

AllanOcelot:
- Added active state to the users icon
- Removed the user (single) icon from containerHeader

### DIFF
--- a/src/components/ContainerHeader.vue
+++ b/src/components/ContainerHeader.vue
@@ -272,7 +272,6 @@ export default {
             }
         },
         nicklistToggle: function nicklistToggle() {
-            console.log(this.sidebarState.sidebarSection);
             if (this.sidebarState.sidebarSection === 'user') {
                 this.sidebarState.close();
             } else {
--- a/src/components/ContainerHeader.vue
+++ b/src/components/ContainerHeader.vue
@@ -30,16 +30,9 @@
                     </a>
                 </div>
                 <div
-                    v-if="sidebarState.sidebarSection === 'user'"
-                    class="kiwi-header-option kiwi-header-option-user kiwi-header-option--active"
-                >
-                    <a @click="sidebarState.close()">
-                        <i class="fa fa-user" aria-hidden="true"/>
-                    </a>
-                </div>
-                <div
                     :class="{
-                        'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist'
+                        'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist',
+                        'kiwi-header-option--active': sidebarState.sidebarSection === 'user'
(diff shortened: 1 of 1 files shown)


src/components/ContainerHeader.vue
```diff
@@ -29,17 +29,10 @@
...
-                >
-                    <a @click="sidebarState.close()">
-                        <i class="fa fa-user" aria-hidden="true"/>
-                    </a>
-                </div>
                 <div
                     :class="{
-                        'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist'
+                        'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist',
+                        'kiwi-header-option--active': sidebarState.sidebarSection === 'user'
```
prawnsalad: You can't have multiple same keys in an object so this should be something like `'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist' || sidebarState.sidebarSection === 'user'`

src/components/ContainerHeader.vue
```diff
@@ -278,6 +271,13 @@ export default {
                 network.ircClient.join(channelName);
             }
         },
+        nicklistToggle: function nicklistToggle() {
+            if (this.sidebarState.sidebarSection === 'user') {
+                this.sidebarState.close();
```
prawnsalad: If it's currently on the userbox then it go back to the nicklist, not close the whole sidebar

src/components/ContainerHeader.vue
```diff
@@ -278,6 +271,13 @@ export default {
                 network.ircClient.join(channelName);
             }
         },
+        nicklistToggle: function nicklistToggle() {
```
prawnsalad: This should use the function shorthand, `nicklistToggle() {`. Also lets rename it to `toggleNicklist()` so it's consistent with the function name a few lines below

AllanOcelot:
@prawnsalad  these have all been fixed, please review when you get the chance :)

src/components/ContainerHeader.vue
```diff
@@ -29,23 +29,16 @@
...
                     :class="{
                         'kiwi-header-option--active': sidebarState.sidebarSection === 'nicklist'
+                            || sidebarState.sidebarSection === 'user'
                     }"
                     class="kiwi-header-option kiwi-header-option-nicklist"
                 >
                     <a
                         :title="$t('person', {count: Object.keys(buffer.users).length})"
-                        @click="sidebarState.toggleNicklist()"
+                        @click="toggleNicklist()"
```
ItsOnlyBinary: Having this as a method seems inconsistent, feel this should be reverted and the method removed.

PR #797, opened 2019-01-24
Away status

ivarmlee11:
Adds a visual indicator for an away status on the nick list and on the message list

### DIFF
--- a/src/components/AwayStatusIndicator.vue
+++ b/src/components/AwayStatusIndicator.vue
@@ -11,12 +11,12 @@
 'kiwi public';
 
 export default {
+    props: ['user'],
     computed: {
         doesNetworkHaveAwayNotifyCap() {
             return this.$state.getActiveNetwork().ircClient.network.cap.isEnabled('away-notify');
         },
     },
-    props: ['user'],
     methods: {
         checkCap() {
             return this.doesNetworkHaveAwayNotifyCap;
--- a/src/components/AwayStatusIndicator.vue
+++ b/src/components/AwayStatusIndicator.vue
@@ -1,9 +1,8 @@
 <template>
     <span
-        v-if="doesNetworkHaveAwayNotifyCap()"
-        :class="{ 'kiwi-awaystatusindicator-away': isUserAway(),
-                  'kiwi-awaystatusindicator-not-away': !isUserAway(),
-                  'kiwi-awaystatusindicator-self': isUserSelf() }"
+        v-if="awayNotifySupported"
+        :class="{ 'kiwi-awaystatusindicator--away': user.isAway(),
+                  'kiwi-awaystatusindicator--self': isUserSelf }"
         class="kiwi-awaystatusindicator"
         @click="toggleSelfAway()"
     />
@@ -15,29 +14,20 @@
 export default {
     props: ['network', 'user'],
     computed: {
-    },
-    methods: {
-        doesNetworkHaveAwayNotifyCap() {
+        isUserSelf() {
+            let user = this.$state.getUser(this.network.id, this.network.nick);
+            return this.user === user;
+        },
+        awayNotifySupported() {
             return this.network.ircClient.network.cap.isEnabled('away-notify');
         },
-        isUserAway() {
-            if (this.user.away) {
-                return true;
-            }
-            return false;
-        },
+    },
+    methods: {
         toggleSelfAway() {
-            if (this.isUserSelf()) {
-                let val = this.isUserAway();
+            if (this.isUserSelf) {
+                let val = this.user.isAway();
                 this.network.ircClient.raw('AWAY', val ? '' : 'Currently away');
             }
-        },
-        isUserSelf() {
-            if (this.user ===
-                this.$state.getUser(this.network.id, this.network.nick)) {
-                return true;
-            }
-            return false;
         },
     },
 };
@@ -54,7 +44,7 @@ export default {
     transition: background 0.2s;
 }
 
-.kiwi-awaystatusindicator-self {
+.kiwi-awaystatusindicator--self {
     cursor: pointer;
 }
 
--- a/src/components/AwayStatusIndicator.vue
+++ b/src/components/AwayStatusIndicator.vue
@@ -1,6 +1,7 @@
 <template>
     <span
         v-if="checkCap()"
+        :user="user"
         :class="{ 'kiwi-awaystatusindicator-away': isUserAway(),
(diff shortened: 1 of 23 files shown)


src/components/MessageListMessageCompact.vue
```diff
@@ -43,6 +43,11 @@
         >
             {{ ml.formatTime(message.time) }}
         </div>
+        <span
+            v-if="message.type === ('privmsg' || 'msg')"
```
prawnsalad: `('privmsg' || 'msg')` this will only ever match against `privmsg`

ivarmlee11: I will address this as well as add a commit that detects whether or not the current network has the 'away-notify' cap.

prawnsalad:
image

The away status should come before the prefix as the prefix is part of the nick

src/components/MessageList.vue
```diff
@@ -338,6 +339,12 @@ export default {
 
             return message.isHighlight;
         },
+        userStatus(user) {
+            if (state.getActiveNetwork().ircClient.network.cap.isEnabled('account-notify')) {
```
prawnsalad: Since this would a rather hot function, moving this `state.getActiveNetwork().ircClient.network.cap.isEnabled('account-notify')` into it's own computed property in this component would cache the result for some performance gains.

Secondly, why is this reliant on account-notify? Should it be away-notify?

src/components/MessageList.vue
```diff
@@ -79,6 +79,7 @@
 <script>
 'kiwi public';
 
+import state from '@/libs/state';
```
prawnsalad: Components can get access to the state via `this.$state` instead of importing it in

src/components/MessageListMessageCompact.vue
```diff
@@ -43,6 +43,11 @@
         >
             {{ ml.formatTime(message.time) }}
         </div>
+        <span
```
prawnsalad: This div is used in a few places. It might be better to create a component for it so it's not duplicated all over the place

AllanOcelot:
Going to start off the morning by fixing prawn's recommendation on this PR - before adding in theme support for the status indicators, and adding the final 'set away status' in the selfUser component.

I'm aiming to have this PR up to scratch and mergable by the end of the day.

ivarmlee11:
I've addressed all of the issues raised except breaking out the status indicator into it's own component. I'll be doing that next.

ivarmlee11:
I created an away status indicator component that can be moved into the self user component as well as the user box component. Slight changes may be needed @AllanOcelot.

AllanOcelot:
@ivarmlee11  - how did you get on with setting the away status via text input in SelfUser?
If you've not had chance to do it, I can easily pick it up - let me know :)

ivarmlee11:
@AllanOcelot I almost have it, there is just some default text that shows when you look at your own away message that I need to do away with

prawnsalad:
Notes from conversation:

~~~
Nick change error message not displaying
Away checkbox to the right, text to just "Away"
Change nick section needs a back button
First section - click status circle to toggle
move user modes after the hostname

PM - show status circle infront of the buffer header
show status circle infront of the nicks on modern view
nicklist - replace the user icon with the away status circle, only if away-notify is enabled
~~~

AllanOcelot:
- When username is submitted, it's going back to the first 'section'
- Add padding to the controlInput text when SelfUser is open, so users can see text
- Animation speeds should all be 0.2s
-  Hostname under username, no divider
- Display checkbox for away on first section
- Wrap Selfuser in controlInput in a vue transition tag
- Cursor:pointer to be added to away Status indicator component
- Check themes colour of away Status Indicator 
- Private message, pass correct user to header component
- Make modern layout follow screenshot design
image
- Themes Updated

AllanOcelot:
- If username is invalid, stay on the second screen of SelfUser
- Need to fix the padding on control input for longer usernames
- Wrap selfUser in vue transtition, check that selfUser is destroyed & created again
- Themes

src/components/AwayStatusIndicator.vue
```diff
@@ -0,0 +1,76 @@
...
+</template>
+
+<script>
+'kiwi public';
+
+export default {
+    props: ['user'],
+    computed: {
+        doesNetworkHaveAwayNotifyCap() {
+            return this.$state.getActiveNetwork().ircClient.network.cap.isEnabled('away-notify');
```
prawnsalad: The component may not always be in relation to the active network. Eg. multiple network listed in the statebrowser. It will need the network object being passed in

src/components/AwayStatusIndicator.vue
```diff
@@ -0,0 +1,76 @@
...
+        },
+    },
+    methods: {
+        checkCap() {
+            return this.doesNetworkHaveAwayNotifyCap;
```
prawnsalad: Why is this simply wrapping the computed property? It could just use the computed property directly...

src/components/AwayStatusIndicator.vue
```diff
@@ -0,0 +1,76 @@
...
+        },
+        isUserAway() {
+            if (this.doesNetworkHaveAwayNotifyCap) {
+                let networkId = this.$state.getActiveNetwork().id;
+                let userToCheck = this.$state.getUser(networkId, this.user.nick);
```
prawnsalad: Don't you already have the user object at `this.user`?

src/components/ContainerHeader.vue
```diff
@@ -215,6 +220,9 @@ export default {
             let content = TextFormatting.styleBlocksToHtml(blocks, showEmoticons, null);
             return content.html;
         },
+        netUser() {
+            return this.$state.getActiveNetwork().ircClient.user;
```
prawnsalad: You will need to be more specific about which network you are getting. If in future we have 2 kiwi windows, both with different ContainerHeader instances, the `getActiveNetwork()` will be incorrect for one of them. You have the `buffer` instance here already, so `buffer.getNetwork()` will always return the correct network

src/components/AwayStatusIndicator.vue
AllanOcelot: Updated, we now check if a network object has been passed in, and if not, we will use the active network.

src/components/AwayStatusIndicator.vue
AllanOcelot: Removed the unneeded function, moved the computed property into the methods and we now use that instead.

PR #802, opened 2019-01-25
Add forcePassword to Welcome.vue startupOptions

ItsOnlyBinary:


prawnsalad:
As mentioned on IRC, using `toggablePassword` instead of `forcePassword` will be less confusing

PR #810, opened 2019-01-31
Remove: Settings icon config from state browser channel

AllanOcelot:
As discussed with @prawnsalad  :) 

- REMOVE: Remove the settings icons from the statebrowser channel list. It's more of an annoyance than a useful feature.


AllanOcelot:
When hovering over a channel, the settings icon is no longer displayed.
image

prawnsalad:
This should now also remove the `showBufferSettings()` method as it's no longer used

PR #814, opened 2019-02-04
Fix notification message sounds

ItsOnlyBinary:


src/libs/AudioManager.js
```diff
@@ -27,7 +27,7 @@ export class AudioManager {
     }
 
     /** Watch the Kiwi state for any message highlights and play an alert */
-    listenForHighlights(state) {
+    listenForMessages(state) {
```
prawnsalad: `watchForMessages`

src/libs/AudioManager.js
```diff
@@ -47,11 +47,11 @@ export class AudioManager {
                 return;
             }
 
-            let isHighlight = message.isHighlight;
+            let isNotification = message.isHighlight || buffer.setting('alert_on') === 'message';
```
prawnsalad: `shouldBleep`

prawnsalad:
Quick addition to this now it actually supports watching for all messages.. it should probably only bleep on message types. Eg. ignoring mode/topic/traffic messages

PR #819, opened 2019-02-06
Experimental: Adding dialog boxes on emit events

AllanOcelot:
Experimental build as we look at adding in new dialog boxes when certain events are emitted.


AllanOcelot:
Main Goals of this PR:

- [ ] Add library for 'Alerts' - some nice notification/action required type modal 
- [ ] Redesign how we filter /  add / channels and users from the stateBrowser 
- [ ] Experiment with the statebrowser design for buttons / events. We can improve on what we have but might have to think outside of the box.

PR #821, opened 2019-02-12
Align channel topic on ModernLayout with other messages.

AllanOcelot:
The topic on modernLayout is now inline with the rest of the Messages, looks much cleaner.


prawnsalad:
Any before / after screenshots of this?

AllanOcelot:
image

image


The topic now lines up on the left, and aligns on the right with the padding of the messageList :)

PR #827, opened 2019-02-14
Away Status Updater Fallback

ivarmlee11:
If a client does not have the 'away-notify' cap this timer updates the away status messages for users in each buffer on a 30 second timer.


AllanOcelot:
Awesome work! - now we have the AwayStatusIndicator in master, it might be worth pulling in master and seeing if it all works together. Nice work and a very good idea :)

ivarmlee11:
@AllanOcelot I'll pull it down and see if it still works as expected.

ivarmlee11:
@AllanOcelot this seems to be working :D

PR #833, opened 2019-02-19
ADD /dice command 

AllanOcelot:
- Add /dice command
- Add autocomplete lookup
- Add translation strings


Shillos:
Shouldn't /dice command be also in aliases too?!

prawnsalad:
/dice and /away are not aliases, they are real commands

PR #849, opened 2019-02-26
Improve input handling on dice command

ItsOnlyBinary:


src/libs/InputHandler.js
```diff
@@ -192,16 +192,12 @@ inputCommands.dice = function inputCommandDice(event, command, line) {
...
-    if (!line.trim()) {
-        rndNumber = Math.floor(Math.random() * sides) + 1;
-    } else {
-        sides = line.replace(/\D/g, '');
-        rndNumber = Math.floor(Math.random() * sides) + 1;
-    }
-    msg = TextFormatting.t('dice_roll', {
+
+    let sides = line.replace(/\D/g, '');
+    sides = typeof sides === 'number' && sides > 0 ? sides : 6;
```
prawnsalad: `sides` will never be of type number here, always a string.

It would also be less confusing to read as:
~~~javascript
sides = parseInt(sides, 10);
if (isNaN(sides) || sides < 0) {
    sides = 6;
}
~~~

PR #851, opened 2019-02-27
Fix media-viewer slot

790:


ItsOnlyBinary:
fixes https://github.com/kiwiirc/plugin-conference/issues/38

PR #858, opened 2019-03-04
Fix audio mentions sometimes not triggering

### LINKED ISSUE #854: Audio nick mentions fail to trigger on certain messages
If the user's nick exists previously in the message, no audio notification happens.

`<bar> want a drink, foo? // works` 

`<bar> want some food, foo? // fails, no audio notification`

The nick is highlighted correctly, so this seems to be an issue solely in the audio notifications.

790:
Fixes #854

### DIFF
--- a/src/helpers/Misc.js
+++ b/src/helpers/Misc.js
@@ -51,7 +51,7 @@ export function splitHost(uri) {
  * @param {string} nick The nick to search for
  */
 export function mentionsNick(input, nick) {
-    if (input.toLowerCase.indexOf(nick.toLowerCase()) === -1) {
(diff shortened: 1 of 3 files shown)


prawnsalad:
Any info on performance regressions here? It originally wasn't using regex since its a hot function

790:
I did run some tests, worse case is same speed (no partial match) and when there's a partial match it's typically been about double the ops/s in Chrome using the regexp, on both long texts and short texts. In Firefox the speeds are about equal.

prawnsalad:
Ok cool. +1 on the the tests

PR #865, opened 2019-03-14
FIX: 862 - Welcome Screen size if recaptcha active

AllanOcelot:
This update simply adds a wider welcome screen form if the user has specific a 'SiteID' for the recaptcha in the config options.

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -1,10 +1,7 @@
 <template>
-    <startup-layout ref="layout" class="kiwi-welcome-simple" >
+    <startup-layout ref="layout" :class="{ 'kiwi-recaptcha-enabled': recaptchaSiteId }" class="kiwi-welcome-simple" >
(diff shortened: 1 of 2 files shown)


AllanOcelot:
Preview: https://imgur.com/a/sCPaNQD

src/components/startups/Welcome.vue
```diff
@@ -1,7 +1,10 @@
 <template>
-    <startup-layout ref="layout" class="kiwi-welcome-simple">
+    <startup-layout ref="layout" class="kiwi-welcome-simple" >
         <template v-slot:connection v-if="!network || network.state === 'disconnected'">
-            <form class="u-form kiwi-welcome-simple-form" @submit.prevent="formSubmit">
+            <form :class="{ 'kiwi-recaptcha-enabled': recaptchaSiteId }"
```
prawnsalad: This class would be cleaner in the top most element just in case anything else needs adjusting in future. ie. `kiwi-welcome-simple--recaptcha`

Issue #842, opened 2019-02-24
Ability to choose notification sound


AllanOcelot:
Very low priority, just whilst it's on my mind:

We should allow our users to choose from a custom sound file (hosted locally) or a drop down menu of 3/4 sound effects for notifications.

This would just add another layer of customisation to Kiwi.

PR #867, opened 2019-03-15
Customaudio2

790:


AllanOcelot:
Some feedback:

- Ideally we would not have the input wrapped in a label, as I normally style the label itself aimed at the file input ( styling file inputs is notoriously difficult ) 
- We need to provide some user feedback, logging the errors to the console is not ideal
- Maybe display some restrictions to the user (max file size, file format etc? )

ItsOnlyBinary:
closes #842

PR #871, opened 2019-03-15
Hide network name on restricted networks

AllanOcelot:
Hide the network name header on restricted networks (as it's the only network)


prawnsalad:
This has removed the add channel box and the link to the server messages

prawnsalad:
Closing this as we really do need the server tab available.

PR #873, opened 2019-03-19
IRCv3 @+draft/typing implementation

790:
Provides support for ircv3 (draft) `typing` message-tag. This requires a version bump of irc-framework before merging (4.2.0 at time of writing).

# Todo

- [x] Display/styling of user's typing state, currently just a placeholder in the nicklist.
- [x] Add settings to disable. Where? Instance operator? global client setting? per channel setting?
- [x] Move MSGTAG into irc-framework rather than dealing with raws?
- [x] Improve multi-channel interaction. Typing states are per target, but currently stored on the user for the whole network. Typing state changes are only processed if the target is the active buffer, but the user flipping between different channels may show an incorrect typing state temporarily.


prawnsalad:
Any reason this needs `performance.now()` instead of just `Date.now()`? Edge nor IE11 supports `performance`

790:
> Any reason this needs `performance.now()` instead of just `Date.now()`? Edge nor IE11 supports `performance`

No particular reason, just what I normally use for time stuff. I checked mdn and it said both were supported but looking at it more, caniuse says no so I'm not sure what the current state of support is. I'll change it to Date.now though for maximum support.

AllanOcelot:
Some ideas on how we could implement this:

prawnsalad:
I've got it working per buffer now, but we still need to decide on the best way to display and position the TypingUsersList component

prawnsalad:
For Sir Allans ref.

image

PR #878, opened 2019-03-27
HOTFIX: Remove height 100% from app.vue line 473.

AllanOcelot:
The bug: When not connected to a channel, in mobile view, the 'disconnected' line at the top is being given a height of 100%. This should not be happening. 

 This bug was previously solved but must have been remerged in recently by some app.vue work.

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -424,7 +424,7 @@ body {
     left: 0;
     width: 200px;
     bottom: 0;
-    transition: all 0.1s;
+    transition: left 0.145s, margin-left 0.145s;
     z-index: 1;
 }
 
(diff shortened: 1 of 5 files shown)


prawnsalad:
This actually shows that the new connected/disconnected bar styling has now overridden the mobile draw overlay.

See the current version on kiwiirc.com/nextclient/ and how when the statebrowser is open on mobile view, the content on the right has a gray overlay. Since the new themes, it is no longer there as you have taken over the .kiwi-workspace::after element for the connection colour.

AllanOcelot:
Looking into this at the moment, I'll look to adding in the overlay shadow if the sidebar is open // try to rework the connecting thing so we can keep both in.

PR #879, opened 2019-03-28
Dark theme update, increase nick brightness

chesty:
The default setting of 40 is unreadable on a black background in my opinion.


AllanOcelot:
@chesty if you have the time and desire, would you want to look at this issue #880  - it's to do with night swatch but it will save us having a separate PR in future for one or two lines of code. :)

chesty:
can you ping me (chesty) next time you're on irc please? cheers.

PR #892, opened 2019-04-11
Connected / Disconnected restyling

AllanOcelot:
I have redesigned the disconnected component that Kiwi displays when a user is disconnected.

Instead of having a large block displaying in the messageList, I have made the component 'sticky' to the top of the message list, allowing the user to clearly see they are disconnected.

( User is disconnected : Note, the changed 'disconnected' messages in the messageList ) 
image

( Connecting ) 
image

( Connected ) 
image

Please provide suggestions and feedback below, thanks!


prawnsalad:
Looking a bit messed up in the server tab

image

AllanOcelot:
Notes:
- Having the 'notConnected' positioned via absolute, with messageList relative, means the 'Not Connected' component would be out of view in long conversations / server tabs.
- Instead the 'not Connected' component is positioned absolute - this hans't cause any bugs in my testing and seems to work very well across the board.
- The icons are now visible on mobile devices too 

image
image

PR #936, opened 2019-04-24
Sidebar styling update

### LINKED ISSUE #922: Make Sidebar's top:0px when on mobile devices
This is a simple fix, but it makes the Kiwi experience feel a bit tighter on mobile devices.

AllanOcelot:
This small PR makes the sidebars for mobile / tablet much better all around with three simple fixes.

- Sidebars for table now sit at the top of Kiwi ( top: -4px )
- Adjusted the lineheight and spacing of the sidebar close section, looks much better now
- Made ALL sidebars the same max width for tablet/mobile devices - feels much better to use now

edit: closes #922

### DIFF
--- a/src/components/Container.vue
+++ b/src/components/Container.vue
@@ -389,7 +389,7 @@ export default {
     }
 
     .kiwi-sidebar {
-        top: 0;
+        top: -4px;
     }
 }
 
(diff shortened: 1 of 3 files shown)


AllanOcelot:
@prawnsalad  - thoughts on this PR?

prawnsalad:
The changing of the max-width here has effected it when embedded in a small page. The sidebar now goes off-screen at times. The reason it was at 100% max-width is so that when it gets into smaller screens/embeds, the content is still viewable even if it does get a bit crammed.

PR #940, opened 2019-04-25
fix error when removing network

ItsOnlyBinary:
```
TypeError: Cannot read property 'isChannel' of null
    at parsedEventsHandler (IrcClient.js?c859:1070)
    at s (browser.js?2f5d:8)
    at i.handle (browser.js?2f5d:8)
    at eval (browser.js?2f5d:8)
    at e.a.emit (browser.js?2f5d:1)
    at emit (browser.js?2f5d:8)
    at eval (browser.js?2f5d:8)
    at e.value (browser.js?2f5d:8)
    at e.value (browser.js?2f5d:8)
    at eval (browser.js?2f5d:8)
```

```
if (_buffer19.isChannel() && !_buffer19.joined) {
    _buffer19.enabled = false;
}
```

### DIFF
--- a/src/libs/IrcClient.js
+++ b/src/libs/IrcClient.js
@@ -1146,6 +1146,11 @@ function clientMiddleware(state, network) {
                 buffer = network.serverBuffer();
             }
 
+            if (!buffer) {
+                // we could not find a buffer, this is likely because the network was removed
+                return;
+            }
+
             // TODO: Some of these errors contain a .error property whcih we can match against,
             // ie. password_mismatch.
 


prawnsalad:
`buffer` should always be a buffer instance. If it's not then there's an issue somewhere else.
https://github.com/kiwiirc/kiwiirc/blob/64a59510071dce2e8b5fb873ffe5ad2b41f9af52/src/libs/IrcClient.js#L1146

Since it happens when removing the network, is the irc error a QUIT message? If so, it can just be added into this if statement as none of that code is needed to be run in that case, https://github.com/kiwiirc/kiwiirc/blob/64a59510071dce2e8b5fb873ffe5ad2b41f9af52/src/libs/IrcClient.js#L1140

PR #946, opened 2019-04-28
Fix typing notifications in direct messages

ItsOnlyBinary:


src/libs/TypingMiddleware.js
```diff
@@ -17,8 +17,11 @@ export default function typingMiddleware() {
             return;
         }
 
+        // if we are params[0] then the target is the sender (direct message)
+        let target = (message.params[0] === client.user.nick) ? message.nick : message.params[0];
```
prawnsalad: This should make case insensitive and use the ternary syntax style we use everywhere else for ease of reading:
~~~javascript
let target = (message.params[0].toLowerCase() === client.user.nick.toLowerCase()) ?
    message.nick :
    message.params[0];
~~~

PR #950, opened 2019-04-30
Move MessageListAvatar Colours to Theme Files

AllanOcelot:
This moves the colour declaration for the messageListAvatar from the component to the theme file.

image


AllanOcelot:
Unsure why these previous commits are appearing here.

image

PR #962, opened 2019-05-07
Smarter banning masks

### LINKED ISSUE #952: Smarter default ban masks
When banning a user:
* If the user is registered, ban their account name.
* If the users hostmask contains an IP, but not solely an IP, ban `*@*ip*`
* If the users username is 9 character long hex, ban `username@*` (ip_as_usernamee format)
* Finally, ban the entire hostmask

This should improve bans around most networks covering different user mask formats

ItsOnlyBinary:
closes #952


src/components/UserBox.vue
```diff
@@ -286,6 +287,30 @@ export default {
             this.network.ircClient.raw('KICK', this.buffer.name, this.user.nick, reason);
         },
         createBanMask: function banMask() {
+            // if the account name is in the host ban the host
+            let accTest = new RegExp('^(.*' + this.user.account + '.*)$');
```
790: While it's unlikely to be a problem, it's probably good practice to `_.escapeRegExp` the account when building the regex.

prawnsalad:
The account name banning can be improved in here. If there is an account name available and the network has an EXTBAN isupport, you can get the account name EXTBAN prefix to specifically ban the account without relying on hostnames. Eg. on inspircd $a is the account EXTBAN prefix.

ItsOnlyBinary:
Now requires #687 to be merged before it will work

PR #967, opened 2019-05-09
Some BOUNCER fixes

DanielOaks:
I've had trouble getting this to work from scratch, so trying to resolve it and keeping my 'fixes' in here.

Issues which this *should* resolve:

- Network list doesn't get populated on a new connection.
- When adding a network, the first non-controller network is named `bnc` (when it should be named `Network1`). However, any networks added after that are correctly named `Network[2,3,etc]` so it's somewhat working.
- Send the password correctly when connecting actual networks other than the controller network.
- The bouncer gets a network saved to it called `bnc` where... instead of the host and port being an upstream network to connect to, the host and port are the bouncer's own details (obv the bouncer can't/shouldn't connect to itself).

Issues I've yet to resolve:

- n/a


DanielOaks:
Alright so as-is, it... mostly works, seemingly? It feels unstable still, and a bit hacky in terms of working, but the net list does get populated, the networks are named appropriately, and the password sending and actual connecting does all work fine. Whoo!

src/libs/BouncerProvider.js
```diff
@@ -196,6 +199,11 @@ export default class BouncerProvider {
         }
 
         this.state.networks.forEach((network) => {
+            // don't save the controller network or add it to the bnc
+            if (controller === network) {
```
DanielOaks: Alright so the issue with this method is that the controller network can **either** be a special separate network (e.g. when you first join, which we don't want to save) or it can be a regular bouncer network that's connected to an upstream network (which we do want to save).

What this should do is check the `BOUNCER` isupport token for a value (but I need to make kiwibnc send that earlier so this check is more accurate and happens at the right time).

prawnsalad: There's a current todo item in the bnc for including custom isupport tokens in a similar fashion to caps, but I'm not sure how that would effect this method in BouncerProvider?

Also, what's this special network thing?

DanielOaks: tl;dr If the controller network is not connected to an upstream (e.g. the automatic controller network that's created when we first join KiwiIRC), then that connection should not be saved to the bouncer. If it does get saved to the bouncer, the bnc has a new network added to it whose details are "Connect to <bouncer_address> on <bouncer_port>" and that's it, which doesn't work and is why we try to avoid this case.

If the controller network is connected to an upstream, then we can save it here without any issues. This piece of code is fixed by the latest commit, and depends on the improved `005` sending in this bouncer PR: https://github.com/kiwiirc/kiwibnc/pull/11

prawnsalad:
If you update this PR from master, there is now a ircClient.bnc.hasNetwork() helper that checks if the BOUNCER 005 has a network tag

DanielOaks:
Okay, I'm fairly sure this is right to be looked over now. Lots of weird, intricate little step-by-step fixes but it acts good now on my end from a brand new connection.

When making a new network, you need to press connect, wait a few seconds and then try again (to let it `saveState()` again and actually add the network.

prawnsalad:
After some testing, it's decided that adding a `hidden` flag to a NetworkState instance so that the initial bouncer controller can hide itself from the statebrowser will be easier going forward.

This will hide the controller network and also open the functionality of plugins to use network connections as other type of controllers in future.

PR #969, opened 2019-05-09
Update to Colour Picker + Emoji picker

AllanOcelot:
This PR updates both the Colour Picker and the Emoji picker to look and function more inline with modern Kiwi standards.


src/components/inputtools/TextStyle.vue
```diff
@@ -1,40 +1,36 @@
...
+                     :key="n"
+                     :class="['irc-bg-colour-' + n]"
+                     :data-code="[n]"
+                     class="kiwi-inputtools-colours-colour"
+                     @click="onColourClick" />
+            </div>
+
+            <div v-else
+                 class="kiwi-inputtools-colours">
+                <div v-for="n in 96"
```
DanielOaks: Just curious, is this meant to be 96 or 98? 98 is where the extended list of colours stops.

AllanOcelot: 98, I updated it ahah

AllanOcelot:
TODO:
- Make plus icon smaller / thinner
- Animate between the two classic/extended colour schemes ( Preferably slide right / left )
- Make borders for both colour and emoji picker thiner
- Ensure borders + content is themed up for each theme

prawnsalad:
Isn't this removing the text colour schemes for each theme that we spent time going over previously?

src/components/inputtools/TextStyle.vue
```diff
@@ -1,40 +1,35 @@
...
-            <div class="kiwi-inputtools-colours-colour irc-bg-colour-pink"
-                 data-code="13" @click="onColourClick"/>
-            <div class="kiwi-inputtools-colours-colour irc-bg-colour-grey"
-                 data-code="14" @click="onColourClick"/>
-            <div class="kiwi-inputtools-colours-colour irc-bg-colour-light-grey"
-                 data-code="15" @click="onColourClick"/>
-            <div class="kiwi-inputtools-colours-colour kiwi-inputtools-colours-reset"
-                 @click="onResetClick"><i class="fa fa-ban" aria-hidden="true"/></div>
+            <div class="kiwi-inputtool-colour-actions" >
+                <div class="kiwi-inputtool-button" @click="showModernPalette = !showModernPalette">
```
prawnsalad: `kiwi-inputtool-button` would be more aptly named `kiwi-inputtool-colour-action` here

src/components/inputtools/TextStyle.vue
```diff
@@ -1,40 +1,35 @@
...
+                    <i v-else class="fa fa-chevron-up" aria-hidden="true" />
+                </div>
+            </div>
+
+            <div class="kiwi-inputtools-colour-container">
+                <div :class="{active: !showModernPalette}" class="kiwi-inputtools-colours">
+                    <div v-for="n in 15"
+                         :key="n"
+                         :class="['irc-bg-colour-' + n]"
+                         :data-code="[n]"
```
prawnsalad: Any reason this value is in an array now?

src/components/inputtools/TextStyle.vue
```diff
@@ -76,33 +72,106 @@ export default {
 };
 </script>
 
-<style lang="less">
+<style lang="less" scroped>
```
prawnsalad: `scroped`

Also, we can remove this less language

src/components/inputtools/TextStyle.vue
AllanOcelot: It's simply looping over the N value - the 'array' brackets are what google said was best practice for concating a predefined string with a looping var :)  

Not sure if :class="{irc-bg-colour + 'n'}" was valid, so turned to google

AllanOcelot:
@prawnsalad  - yes, this removes the theme overrides for colours, as we are essentially quadrupling the colours we have, made more sense to me to have the 98, giving our users more choice to use an appropriate colour

prawnsalad:
We tweaked the first set of colours for each theme so that they better suited. ie. yellow on the default theme was slightly darker so its readable.
The 16-98 colours are the only ones that should never change between themes.

MrGrays:
i am testing this PR at the moment and when you type just numbers it does not show the first number for example if i did 1111 it would only show 111, We have checked this out on all themes on our server and its the same issue. Have checked this out on the builds site too with same issue

marthellius:
Are there any plans to merge the changes proposed by AllanOcelot?

More than 4 years have passed and we are still limited to the 16 color palette from the previous century. 

Would it be possible to implement the new color picker as a plugin, so that the users needing it can actually use it (independently of what happens with this PR)?

clort81:
I have a new 9025 color palette in the wings which will be better.

PR #973, opened 2019-05-12
api addUi userbox_button

### LINKED ISSUE #897: API allow adding buttons to userbox
it would be very handy if plugins could add buttons to the userbox to allow things like linking to report user form etc

ItsOnlyBinary:
not sure if the css for this is going to need looking over by allan

Test plugin here: https://gist.github.com/ItsOnlyBinary/cb0b1d788b752485d093d1d20c92215b

closes #897

### DIFF
--- a/src/components/UserBox.vue
+++ b/src/components/UserBox.vue
@@ -32,7 +32,7 @@
                 <i class="fa fa-question-circle" aria-hidden="true" />
                 {{ $t('more_information') }}
             </a>
-            <div class="kiwi-userbox-plugin-actions">
+            <div class="kiwi-userbox-actions kiwi-userbox-plugin-actions">
                 <div
                     v-for="plugin in pluginUiButtonElements"
                     :key="plugin.id"
@@ -45,7 +45,7 @@
                             }
                         }
                     }"
-                    class="kiwi-userbox-plugin-action"
+                    class="kiwi-userbox-action"
                 />
             </div>
         </div>
@@ -471,6 +471,7 @@ export default {
     opacity: 0.6;
     cursor: default;
     word-break: break-all;
+    padding-left: 1px;
 }
 
 .fa-user.kiwi-userbox-icon {
@@ -528,6 +529,10 @@ export default {
         border-radius: 3px;
     }
 
+    .kiwi-userbox-action:empty {
+        display: none;
+    }
+
     label {
         display: block;
         cursor: pointer;
@@ -540,13 +545,7 @@ export default {
 }
 
 .kiwi-userbox-plugin-actions {
-    display: flex;
-    flex-wrap: wrap;
-    justify-content: center;
-}
-
-.kiwi-userbox-actions:empty {
-    padding: 0.5em;
+    padding: 0;
 }
 
 .kiwi-userbox-opactions {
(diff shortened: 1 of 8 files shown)


AllanOcelot:
This adds an easy way to add plugin buttons to the userBox I take it?

static/themes/common/base.css
```diff
@@ -320,7 +320,8 @@
     color: var(--brand-default-fg);
 }
 
-.kiwi-userbox .kiwi-userbox-actions .kiwi-userbox-action:hover {
+.kiwi-userbox .kiwi-userbox-actions .kiwi-userbox-action:hover,
```
prawnsalad: This line can simply be `.kiwi-userbox-action:hover,`

AllanOcelot: This has been made less specific now

src/components/UserBox.vue
```diff
@@ -402,8 +413,10 @@ export default {
     padding: 1em;
     text-align: center;
     margin: 0;
+    user-select: none;
     box-sizing: border-box;
 
+    .kiwi-userbox-plugin-actions a,
```
prawnsalad: What is this selector referring to?

static/themes/common/base.css
```diff
@@ -315,12 +315,13 @@
...
 
-.kiwi-userbox .kiwi-userbox-actions .kiwi-userbox-action {
+.kiwi-userbox-action {
     border: 1px solid var(--brand-default-fg);
     color: var(--brand-default-fg);
 }
 
-.kiwi-userbox .kiwi-userbox-actions .kiwi-userbox-action:hover {
+.kiwi-userbox-action:hover,
+.kiwi-userbox-plugin-actions a:hover {
```
prawnsalad: Should this be `.kiwi-userbox-plugin-action:hover`?

src/components/UserBox.vue
AllanOcelot: This is styled because plugins may add simple links in, instead of always using 'kiwi-userbox-plugin-action' class :)

static/themes/common/base.css
AllanOcelot: This is for the same reason as above :)

prawnsalad:
This could do with passing some props to the plugin like we do here https://github.com/kiwiirc/kiwiirc/blob/master/src/components/ControlInput.vue#L97

Passing `{el: plugin.el, props: { user: user, userbox: self  }}` where `self` is a computed property returning `this`

ItsOnlyBinary:
this doesnt appear to work with using data
typeof this.$el.dataset.user === "string" in my test plugin

ItsOnlyBinary:
As a solution i propose we use:
```
props: {
    kiwiParent: self,
}
```
this should help prevent any conflicts and also keep it consistent throughout the api.
this would mean for UserBox.vue the user object could be accessed as follows:
```
this.$el.kiwiParent.user
```

prawnsalad:
Actually that makes sense, but I would use a `.kiwi` property to keep them separate and easier for plugins to see what they have.
```
props: {
    kiwi: { user: this, etc: etc }
}
```

ItsOnlyBinary:
Note ive changed ControlInput.vue to match the others which will be a breaking change for anyone using the old props

ItsOnlyBinary:
rebased

ItsOnlyBinary:
rebased

PR #975, opened 2019-05-14
Move StateBrowser Usermenu into own component

AllanOcelot:
As part of an effort to keep Kiwi modular, the StateBrowser Usermenu has been moved to it's own component.


prawnsalad:
This needs the network object being passed to it via a prop so it's a little more loosely coupled

PR #978, opened 2019-05-15
Updates to the control input tools / send button

AllanOcelot:
This PR tweaks the icons in the bottom right corner of Kiwi.

For mobile, the send button has been repositioned to be the furthest right side, this is more traditional and should feel more natural to mobile users.

The Icon for 'plugins' has been changed from bars to ellipsis, along with a close button when opened.

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -54,10 +54,6 @@
                         @keyup="inputKeyUp($event)"
                         @click="closeInputTool"/>
                 </div>
-                <button
-                    v-if="shouldShowSendButton"
-                    type="submit"
-                    class="kiwi-controlinput-send fa fa-paper-plane" />
             </form>
 
             <div ref="plugins" class="kiwi-controlinput-tools">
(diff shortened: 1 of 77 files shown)


src/components/ControlInput.vue
```diff
@@ -93,7 +82,22 @@
...
+                    class="kiwi-controlinput-tools-container-expand"
+                    @click="showPlugins=!showPlugins"
+                >
+                    <i v-if="!showPlugins" class="fa fa-ellipsis-h" aria-hidden="true"/>
+                    <i v-else class="fa fa-caret-right" aria-hidden="true" />
+                </div>
             </div>
+
+            <button v-if="shouldShowSendButton"
+                    class="kiwi-controlinput-send-button"
```
prawnsalad: You've added the new `kiwi-controlinput-send-button` class but the existing `kiwi-controlinput-send` classes still exist and even modified.

It would be better to keep `kiwi-controlinput-send` instead just incase anyone has plugins making use of `kiwi-controlinput-send` already. Don't forget to remove the obsolete CSS styles too

AllanOcelot: I didn't see this change request, shall do first thing Monday morning!

PR #987, opened 2019-05-18
Fix lint error

pjc0247:


prawnsalad:
This lint error happens when you use npm to install dependancies instead of yarn. The build steps in the readme mentions the correct way to install.

PR #999, opened 2019-05-21
Hotfix stop topics being set empty

AllanOcelot:
This simple PR adds a check to see if the user is attempting to set the topic as ''.


ItsOnlyBinary:
I dont feel this is the solution, as people might want to remove the topic completely (it seems the way to do that is to set it as a space)

as an alternative solution i propose #1005

AllanOcelot:
I agree, much bettter

PR #1012, opened 2019-05-24
Don't send +draft/typing done after sending a message, instead use the incoming message to remove the user from typing list

ItsOnlyBinary:


src/components/ControlInput.vue
```diff
@@ -465,7 +465,10 @@ export default {
         },
         submitForm() {
             let rawInput = this.$refs.input.getValue();
-            if (!rawInput) {
+
+            // dont send just whitespace messages clear the input instead
+            if (!rawInput.trim()) {
```
prawnsalad: I'm not sure we should be doing this. Empty lines are commonly used during conversation/jokes/ascii/etc and that's not something we should be preventing.

src/components/ControlInput.vue
```diff
@@ -583,7 +586,7 @@ export default {
 
             this.lastTypingTime = Date.now();
         },
-        stopTyping() {
+        stopTyping(messageSent) {
```
prawnsalad: `messageSent` would be better made more generic, something like `sendStopPause`. There may be a case where we also don't want to send stop/pause that isn't related to sending a message.

src/libs/TypingMiddleware.js
```diff
@@ -22,12 +25,17 @@ export default function typingMiddleware() {
             message.nick :
             message.params[0];
 
+        // if its a privmsg without typing tag emit done
+        let status = (message.tags['+draft/typing']) ?
```
prawnsalad: This would be simpler read as just `let status = message.tags['+draft/typing'] || 'done';`

PR #1019, opened 2019-05-28
Messagelist inline timestamp default colour

AllanOcelot:
Inline timestamps are meant to show on hover, but in Master, on a dark theme these are hard to see.

This PR simply makes it so the colour should always contrast to the themes background, making inline timestamps visible.


AllanOcelot:
Closing in favour of new PR

Issue #1016, opened 2019-05-27
Type notify container alignment on mobile


H7-25:
On mobile type notify container should be aligned to left like messages.
![Schermata 2019-05-27 alle 22 10 54](https://user-images.githubusercontent.com/2895452/58437978-b8958080-80cc-11e9-9e97-743b27859b44.png)

H7-25:
If there's a lot of talking people it also go over text input

H7-25:
Also on large screen if there's a lot talking people it go over userlist, thann over textinput

AllanOcelot:
@H7-25 I will look into this over the next couple of days and try to fix the issues, make this look nicer.

PR #1023, opened 2019-05-30
Update the styling of the typing user list

AllanOcelot:
Should help with #1016 

- Remove background from the typing notification container
- Position the typing container flush left with the messageList
- Added padding to accommodate for a huge typing list of users
- Align left stops the users from simply centering on mobile devices

image

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -837,6 +837,10 @@ export default {
 .kiwi-typinguserslist {
     position: absolute;
     top: -24px;
-    background: var(--brand-default-bg);
+    width: 100%;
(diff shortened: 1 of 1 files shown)


prawnsalad:
The left alignment was needed here so that it's in line with the message input. I think all this needs is the text alignment and the overflow probably.

Once the CSS grid layout is in place, that will solve a lot of the issues in a simpler way too.

AllanOcelot:
I've removed the other properties - but have not tested this yet, so please don't merge.

My understanding if that once we have the CSS grid overhaul, we will revisit this bug?

Issue #986, opened 2019-05-17
[Mobile] Add an option, like some apps already do, to display password in plaintext


TheMythPT:
This is what I said in IRC and I think it's quite explanatory:
> so, I was thinking in something: when using mobile, it's easy to make mistakes typing the password on the password field for SASL authentication. I was wondering if would make sense to have a "display password" thing?

AllanOcelot:
Awaiting PR merge, great suggestion, thanks @TheMythPT  :)

TheMythPT:
Great! Thanks for the hard work!!!

PR #1024, opened 2019-05-30
Add 'View plaintext password' to welcome startup screen

AllanOcelot:
This looks at adding in the suggestion of #986 

We now have an additional checkbox under the password field that will toggle between a password field and a standard text field for the user.

image

image

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -33,10 +33,17 @@
                     v-focus
                     v-if="showPass && (show_password_box || !toggablePass)"
                     :label="$t('password')"
+                    :type="showPlainText ? 'text' : 'password'"
                     v-model="password"
                     class="kiwi-welcome-simple-password u-input-text--reveal-value"
-                    type="password"
                 />
+                <label
+                    v-if="showPass && (show_password_box || !toggablePass)"
+                    class="kiwi-welcome-simple-have-password"
+                >
+                    <input v-model="showPlainText" type="checkbox" >
+                    <span> Display password as plaintext? </span>
+                </label>
                 <input-text
                     v-if="showChannel"
                     :label="$t('channel')"
@@ -94,6 +101,7 @@ export default {
             recaptchaSiteId: '',
             recaptchaResponseCache: '',
             connectWithoutChannel: false,
+            showPlainText: false,
         };
     },
     computed: {
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -29,23 +29,16 @@
                     <input v-model="show_password_box" type="checkbox" >
                     <span> {{ $t('password_have') }} </span>
                 </label>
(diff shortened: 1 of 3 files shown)


AllanOcelot:
Updated Images:
image

image

prawnsalad:
The icon should probably be inline with the text? And the text area not going under the icon. Using flexbox for the input+icon might be a lot easier here

image

src/components/startups/Welcome.vue
```diff
@@ -94,6 +97,7 @@ export default {
             recaptchaSiteId: '',
             recaptchaResponseCache: '',
             connectWithoutChannel: false,
+            showPlainText: false,
```
prawnsalad: Is this property needed anymore?

src/components/utils/InputText.vue
```diff
@@ -144,4 +161,27 @@ export default Vue.component('input-text', {
...
+    cursor: pointer;
+    opacity: 0.5;
+    transition: opacity 0.2s;
+}
+
+.kiwi-show-plaintext.active,
+.kiwi-show-plaintext:hover {
+    opacity: 1;
+}
+
```
prawnsalad: All of the above class names should be using the `u-input-text-` prefix

src/components/utils/InputText.vue
```diff
@@ -68,6 +81,9 @@ export default Vue.component('input-text', {
         updateValue: function updateValue(newValue) {
             this.$emit('input', newValue);
         },
+        checkIfEdge() {
```
prawnsalad: This would be better as a computed property called `isEdgeBrowser` to be more clear

src/components/startups/Welcome.vue
AllanOcelot: This has been removed, unused :)

src/components/utils/InputText.vue
```diff
@@ -144,4 +161,27 @@ export default Vue.component('input-text', {
...
+    cursor: pointer;
+    opacity: 0.5;
+    transition: opacity 0.2s;
+}
+
+.u-input-plaintext.active,
+.u-input-plaintext:hover {
+    opacity: 1;
+}
+
```
prawnsalad: All of the above class names should be using the `u-input-text-` prefix, eg. `.u-input-text-plaintext`.

src/components/utils/InputText.vue
```diff
@@ -144,4 +161,27 @@ export default Vue.component('input-text', {
...
+    display: inline-block;
+    line-height: 40px;
+    width: 30px;
+    text-align: center;
+    cursor: pointer;
+    opacity: 0.5;
+    transition: opacity 0.2s;
+}
+
+.u-input-plaintext.active,
```
prawnsalad: `.u-input-plaintext.active` should be using a modifier `.u-input-plaintext--active`

PR #1039, opened 2019-06-05
Fix to display timestamps on Inline style correctly

westor7:
Currently it doesn't display the timestamps correctly even if the "Show timestamps" option is enabled, i tested by removing the "display: none" and is working perfectly and now it listening the option "Show timestamps" correctly.


AllanOcelot:
ah, this is actually intended behaviour of this layout.

This layout is meant for video streams / smaller embedded versions of Kiwi, where space is limited, so the time stamp only appears on hover.

PR #1040, opened 2019-06-05
Only show errors in network settings if we are not registered

### LINKED ISSUE #883: Errors leaking into server details
seems all errors when caught are added to server details tab, even when not relevant to connecting

image

ItsOnlyBinary:
I'm not sure this is the perfect solution but is cleaner than the current situation.

seems overkill to create more complex code to detect if it was attempting to connect at the time of the error

closes #883


prawnsalad:
Hm yea this would break down since some errors don't close the connection, so the error message would then be hidden.

It should be fairly easy to add a `if (isRegistered) { network.last_error = ; }` wrapper around where we set the error.

ItsOnlyBinary:
@prawnsalad this is fixed and forced pushed ;)

PR #1047, opened 2019-06-12
Fix auto complete replacing words when no replacement is selected

ItsOnlyBinary:
to reproduce this issue start typing something that will not complete to a nick, press tab (invisibility opens an empty auto complete list) then press space. The text you started typing will be replace by emptiness

also typing a none auto completing word press tab then try continue typing

### DIFF
--- a/src/components/AutoComplete.vue
+++ b/src/components/AutoComplete.vue
@@ -36,11 +36,10 @@
 'kiwi public';
 
 import _ from 'lodash';
-import state from '@/libs/state';
 
 export default {
     props: ['filter', 'buffer', 'items', 'close_empty'],
-    data: function data() {
+    data() {
         return {
             items_: [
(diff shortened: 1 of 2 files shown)


790:
There's a small visual distortion as the autocomplete still appears, but empty. Another way to fix might be to check filteredItems and have the autocomplete close itself on mount if there's nothing to display.

ItsOnlyBinary:
that is somewhat intentional, as its desired that on backspace auto complete would start showing options (when triggered with #/@), i will however look into closing auto complete when its triggered by tab if there is no results (if thats possible)

ItsOnlyBinary:
AutoComplete.vue could do with a tidy up, i will add that on sunday

ItsOnlyBinary:
empty auto complete can become populated when a user start to remove chars.

closed in favour of #1479

PR #1050, opened 2019-06-12
SidebarAboutBuffer highlights improvements

ItsOnlyBinary:
Makes the messages html instead of unformatted text
allows you to click the highlight to scroll to the message in the message list

### DIFF
--- a/src/components/MessageList.vue
+++ b/src/components/MessageList.vue
@@ -249,6 +249,12 @@ export default {
         this.listen(this.$state, 'mediaviewer.opened', () => {
             this.$nextTick(this.maybeScrollToBottom.apply(this));
         });
+
+        this.listen(this.$state, 'messagelist.scrollto', (opt) => {
+            if (opt.id) {
+                this.maybeScrollToId(opt.id);
+            }
+        });
     },
     methods: {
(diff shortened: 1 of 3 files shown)


src/components/MessageList.vue
```diff
@@ -421,6 +427,16 @@ export default {
                 this.$el.scrollTop = this.$el.scrollHeight;
             }
         },
+        maybeScrollToId(id) {
+            let msgs = this.$el.getElementsByClassName('kiwi-messagelist-message');
```
prawnsalad: This would be easier using `this.$el.querySelector('.kiwi-messagelist-message[data-message-id="'+theid+'"]`

src/components/MessageList.vue
```diff
@@ -249,6 +249,12 @@ export default {
         this.listen(this.$state, 'mediaviewer.opened', () => {
             this.$nextTick(this.maybeScrollToBottom.apply(this));
         });
+
+        this.listen(this.$state, 'messagelist.scrollto', (opt) => {
+            if (opt.id) {
```
prawnsalad: This should probably be `if (opt && opt.id) {`

PR #1062, opened 2019-06-26
Allow forcing polling mode from config or uri param

ItsOnlyBinary:


src/libs/IrcClient.js
```diff
@@ -42,13 +42,20 @@ export function create(state, network) {
...
         ircClient.options.username = network.username || network.connection.nick;
         ircClient.options.gecos = network.gecos || 'https://kiwiirc.com/';
         ircClient.options.encoding = network.connection.encoding;
 
         state.$emit('network.connecting', { network });
 
+        // Pass socksjs options so we can force polling mode
+        let socksOptions = {};
+        if (ircClient.options.poll) {
+            socksOptions.transports = ['xhr-polling', 'iframe-xhr-polling', 'jsonp-polling'];
```
790: Is the goal of this PR to specifically enable the use of polling or to exclude the use of websockets? If the latter, I'd suggest we prefer `eventsource` / `iframe-eventsource` (http server-sent events), or even `xhr-streaming` over the polling transports where possible to keep resource usage lower.

ItsOnlyBinary: The plan is to prevent the use of websockets, for people on strict networks where websockets can open but cannot be sustained due to being killed by firewalls

PR #1066, opened 2019-06-27
Fix: Remove nickname hover dark

AllanOcelot:
A quick fix that removes the background colour when hovering over a nickname in the messagelist. Most noticeable on the dark theme.


AllanOcelot:
The issue is, the background colour was needed for long nicknames on hover.
We need to have a think about how to handle the hiding and showing of long nicks on hover, maybe css isn't the best answer.

PR #1069, opened 2019-06-27
polyfill's for vue-virtual-scroller & ie11

ItsOnlyBinary:


790:
vue-virtual-scroller in IE11 also requires the `'core-js/fn/array/virtual/includes'` polyfill.

ItsOnlyBinary:
are you sure its not String.includes? thats what got dev server working in ie11 but didnt seem related to virtual scroller

790:
`validator: (value) => ['vertical', 'horizontal'].includes(value)`

Is the code from vue-virtual-scroller that was throwing the error. Happened when trying to /join a new channel from the server buffer.

There's also an error in IE11 where we use `String.startsWith`. There's only 1 instance of this use so maybe not worth adding it to the polyfills and we can change it to use indexOf

ItsOnlyBinary:
it was caching tricking me into thinking it was string.includes it would seem, i just noticed the startsWith issue myself, so have added a fix in this pr for it (will be pushing shortly)

PR #1072, opened 2019-06-30
Respond correctly to bnc sending 'disconnected' state on a network

DanielOaks:
Forgot to open this up last week. Pretty simple change, just makes Kiwi respect when a bouncer sends a disconnected state to us (so it doesn't sit there forever in the 'waiting' state.

### DIFF
--- a/src/libs/BouncerMiddleware.js
+++ b/src/libs/BouncerMiddleware.js
@@ -61,6 +61,14 @@ export default function bouncerMiddleware() {
             });
         }
 
+        if (params[0] === 'state' && params[2] === 'disconnected') {
+            let netName = (params[1] || '').toLowerCase();
+            let eventObj = {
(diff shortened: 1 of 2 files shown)


prawnsalad:
Got an example on how to reproduce this waiting state? Currently this PR will make kiwi think the connection to the bouncer as been closed when it's actually still active. Also, bare in mind that the bouncer state command can be relating to any network in the users account.

DanielOaks:
Hmm, yeah I'm not entirely sure why this was working the way it was before. I'm gonna keep working with this and letcha know once it's ready for review.

prawnsalad:
Closing this as the connection is now entirely closed upon state disconnected. https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/BouncerProvider.js#L113

PR #1082, opened 2019-07-07
Set server buffer as active when no channels are joined at startup

### LINKED ISSUE #1081: connectWithoutChannel does not display server tab
When using startupOptions.connectWithoutChannel: true, it does not display server tab, but have to manually click it.

ItsOnlyBinary:
closes #1081

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -306,6 +306,11 @@ export default {
                 }
             });
 
+            // switch to server buffer if no channels are joined
+            if (!hasSwitchedActiveBuffer) {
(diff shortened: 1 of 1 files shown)


src/components/startups/Welcome.vue
```diff
@@ -306,6 +306,11 @@ export default {
                 }
             });
 
+            // switch to server buffer if no channels are joined
+            if (!hasSwitchedActiveBuffer) {
+                state.setActiveBuffer(net.id, '*');
```
prawnsalad: `net.serverBuffer().name` would be preferred over `*` here for readability

ItsOnlyBinary: forced pushed, thanks

PR #1098, opened 2019-07-09
Font color improvements for network channel unread message count.

TheBeerbarian:
Dark and Elite these have unreadable message counts. This PR makes subtle changes to use the themes default background variable.

image


TheBeerbarian:
Corrected after picture of dark theme.
image

PR #1101, opened 2019-07-10
New Welcome screen

AllanOcelot:
This PR contains a new welcome screen for Kiwi, hoping to eventually lead to us phasing out the existing 'welcome' screen.


westor7:
That new welcome screen is perfect, just need some minor fixes.

Image Preview: https://i.imgur.com/y6ABRXu.jpg

1. From the image above the "Password" doesn't get translated.
2. From the image above it would be nice the eye icon to be little bit more large (10-20%).
3. From this image https://i.imgur.com/wzsmIz2.jpg the loading icon should be more larger.

- Keep rocking like this dude!! <3

AllanOcelot:
Thanks for the feedback @westor7  , I'll edit these in and push them in the morning :) Glad you like it :)

westor7:
Nice work @AllanOcelot , it seems 1 and 2 has been done, only the 3rd item didn't get fixed, the loading icon is still small.

ItsOnlyBinary:
I think the split could do with having a max width, everything starts getting a little wide at 1440p

westor7:
There are some display issues.

The font size is now smaller (i don't know if this is for purpose), but this is not for all fields, the "Channel" text is bigger than "Nickname" same for the channel input field, also the eye icon doesn't get changed when clicked as was previously.

The suggestion is to have the same font size for all the items, and increase the eye icon size little bit, also when someone click on the eye icon to be changed.

westor7:
Nice, it seems they have been fixed, now the only issues are:

1. The loading logo after you click Start is too small (it needs 50% at least more size).
2. There is a bug, if you click on the checkbox "I have a password" and then you enter a password and then you uncheck the checkbox and click on Start button, it doesn't allow you to connect because it take effect the password field you enter, even if you have "I have a password" checkbox uncheck, it should be there an extra if statement check to check if the checkbox is unchecked to skip the connection with the password and connect thought.

digitalcircuit:
It looks like this doesn't update the `Start` button based on nickname?

[With current Kiwi IRC's welcome screen](https://builds.kiwiirc.com/kiwiirc/master/ ), putting in `First.Last` results in the `Start` button being grayed out.  Confusing (*it's not obvious what's wrong, one person assumed you needed a password too*), but at least the button's disabled.

[Testing this PR](https://builds.kiwiirc.com/kiwiirc/pull-1101/ ), it seems like even that feedback is gone, so `Start` with a nickname of `First.Last` just does nothing.  `Start` also doesn't disable if you clear out the nickname entirely.

prawnsalad:
Thanks @westor7 and @digitalcircuit for the feedback, looks like it's all ready to go now \o/

PR #1107, opened 2019-07-14
optionally store nicklist scroll position and restore it after switching buffers

### LINKED ISSUE #1104: Nicklist remember last selected nick
I'm receiving a lot of users feedback asking me to let nicklist remember latest selected nick, so if they select and open an userbox info when they switch back to nicklist they are in the latest selected position and not at the begin of the list again.

ItsOnlyBinary:
this uses vue-virtual-scroller emitUpdate that comes with a warning below, the buffer setting does also disable this emitUpdate feature

`
emitUpdate (default: false): emit a 'update' event each time the virtual scroller content is updated (can impact performance).`

closes #1104

### DIFF
--- a/src/components/Nicklist.vue
+++ b/src/components/Nicklist.vue
@@ -27,12 +27,12 @@
         <DynamicScroller
             ref="scroller"
             :items="sortedUsers"
-            :min-item-size="34"
+            :min-item-size="39"
             :key-field="'nick'"
             :emit-update="storeScrollIndex"
             class="kiwi-nicklist-users"
             @visible="onScrollerVisible"
-            @update="onScrollerUpdate"
+            @update="debounceScrollerUpdate"
         >
             <template v-slot="{ item, index, active }">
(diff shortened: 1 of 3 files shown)


prawnsalad:
This looks to be restoring a little too much of the position. Nick1 being at the top of the nicklist, open the userbox, go back to the nicklist, then Nick1 is in the middle of the nicklist.

ItsOnlyBinary:
Rebased and fixed it to correctly find the first nick scrolled into view.

Also added it emitting an event for the current users in view that do not have avatars (for plugins to load them on demand)

CoryChaplin:
I've tested this and had some issues.
The position seems to be shared between channels that makes it quite confusing. And in some cases after switching around many tabs it ended up reset.
I love the idea but I think it needs some extra tuning.

ItsOnlyBinary:
Yeh with vue-virtual-scroller this is not working as intended. I've even try implementing my own scrollToItem without much luck

PR #1108, opened 2019-07-14
Fix sorting of numbered nicks

ItsOnlyBinary:
eg:
```
testNick1
testNick11
testNick2
```


prawnsalad:
This is a very hot function so applying regex here kills performance.

That said, have you got a screenshot where it's not working already?

ItsOnlyBinary:
now using Intl.Collator

ItsOnlyBinary:
Not working in ie11 :(

PR #1110, opened 2019-07-16
Rename translations msgid to use english translation

ItsOnlyBinary:


prawnsalad:
Closing as we can now use Crowdin correctly using the existing keys and we have now learned that the i18next lib we use would break in all different ways using English as keys.

PR #1115, opened 2019-07-18
HOTFIX: Fixed the 'greetingText' multi-line spaces

westor7:
It now has more space between multi-lines, so now the text looking more pretty.


prawnsalad:
This effects the entire application. Did you just want to effect the startup screen..? If so, that should be done in components/startups/CommonLayout.vue

PR #1116, opened 2019-07-18
HOTFIX: Fixes the 'greetingText' multi-line spaces

westor7:
Startup screen now has more space between multi-lines, so now the text looking more pretty.


westor7:
wrong change!

PR #1121, opened 2019-07-22
Only use Intl.Collator in nicklist if its supported (fixes ios9)

### LINKED ISSUE #1120: White screen on ipad2 ios 9.3.5
On ipad2 with ios 9.3.5 (more recent version available for ipad2) kiwiirc doesn't work and show a white screen.

ItsOnlyBinary:
not sure if there is a neater way to do this, but it does appear to work (dont have ios device to test)

closes #1120

### DIFF
--- a/src/components/Nicklist.vue
+++ b/src/components/Nicklist.vue
@@ -52,10 +52,25 @@ import NicklistUser from './NicklistUser';
 
 let log = Logger.namespace('Nicklist');
 
+// This provides a better sort for numbered nicks but does not work on ios9
+let intlCollator = null;
+if (global.Intl) {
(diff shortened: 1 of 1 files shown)


prawnsalad:
Something a bit cleaner than rewriting functions...

~~~javascript
// This provides a better sort for numbered nicks but does not work on ios9
let intlCompare = null;
if (global.Intl) {
    intlCompare = new Intl.Collator({}, { numeric: true }).compare;
}

function strCompare(a, b) {
    if (intlCompare) {
        return intlCompare(a, b);
    }

    if (a === b) {
        return 0;
    }

    return a > b ?
        1 :
        -1;
};

~~~

ItsOnlyBinary:
Fixed, thanks prawn

PR #1122, opened 2019-07-23
Ignore error messages that match the last message in the buffer

### LINKED ISSUE #996: User spammed when not registered in chan
When attempting to message a channel requiring users to be logged in  / registered, we are spamming the user with 
 >'You need to be identified to a registered account to message this channel'

This appears every time the user tries to type, we should limit the output of this message to one.

ItsOnlyBinary:
closes #996


prawnsalad:
Why would you want to ignore error messages? You could be trying to set a channel mode or do some type of server action while tweaking your command, and you would never know why your command is failing if it doesn't show anything

ItsOnlyBinary:
Doh didnt think of that one, maybe need to make it more targeted towards the error type.

basically when a user is typing in a channel they cannot send to, errors are triggered by typing message-tags resulting in error spam

ItsOnlyBinary:
Rebased and added support for 477 ERR_NEEDREGGEDNICK to go to the correct buffer (with norepeat)

PR #1125, opened 2019-07-24
Dont show nicklist button highlighted when a user is viewed in the sidebar

ItsOnlyBinary:
this makes it clearer that none of the header buttons are currently active


ItsOnlyBinary:
Doesnt quite feel right :(

PR #1133, opened 2019-07-31
Fix multiple clicks on IrcInput in firefox

### LINKED ISSUE #1130: Clicking already focused inputbar clears focus (Firefox)
kiwi-focus

In the screen recording, I am just left clicking repeatedly on the input bar.

`Mozilla/5.0 (X11; Linux x86_64; rv:69.0) Gecko/20100101 Firefox/69.0`

Does not happen in Chromium.

ItsOnlyBinary:
closes #1130

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -747,7 +747,7 @@ export default {
     height: 100%;
     box-sizing: border-box;
     overflow: visible;
-    padding-top: 8px;
+    padding: 7px 0 12px 0;
 }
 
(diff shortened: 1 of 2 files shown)


prawnsalad:
Replacing these lines:
~~~
    min-height: 1.6em;
    max-height: 4.8em;
~~~

with a simple `height: 100%;` has the same fixed effect while being cleaner. I *think* you mentioned you had tried the 100% height... any reason that's not in use here?

PR #1166, opened 2019-10-03
Add getLoadingState() function to BufferState.js

tralves:
The buffer takes some time to load the chat history. We need a way to provide the UI with the information that the buffer is still loading those messages.

This PR adds the function `getLoadingState()` that returns one of the following states:

- `disconnected`: network is disconnected;
- `connecting`: network is connecting;
- `loading`: buffer is loading the chat history;
- `done`: buffer has finished loading and is now ready;

It adds also the function `isReady()` which just returns true if the LoadingState is `done`.

### DIFF
--- a/src/libs/state/BufferState.js
+++ b/src/libs/state/BufferState.js
@@ -63,8 +63,24 @@ export default class BufferState {
             maybeStartWhoLoop(this);
         }
 
-        state.$on('network.connecting', this.onNetworkConnecting.bind(this));
-        state.$on('buffer.close', this.onBufferClose.bind(this));
+        function onNetworkConnecting(event) {
+            if (event.network === this.getNetwork()) {
+                this.chathistory_request_count = 0;
+            }
+        }
+
+        function onBufferClose(event) {
+            if (event.buffer === this) {
+                this.state.$off('network.connecting', onNetworkConnectingBound);
+                this.state.$off('buffer.close', onBufferCloseBound);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/libs/state/BufferState.js
```diff
@@ -396,6 +404,46 @@ export default class BufferState {
...
+    onNetworkConnecting(event) {
+        if (event.network === this.getNetwork()) {
+            this.chathistory_request_count = 0;
+        }
+    }
+
+    onBufferClose(event) {
+        if (event.buffer === this) {
+            this.state.$off('network.connecting', this.onNetworkConnecting);
+            this.state.$off('buffer.close', this.onBufferClose);
```
prawnsalad: These 2 $off calls won't do anything as the event listeners are attached using `.bind()` which returns a new function, not these specific ones being removed

src/libs/state/BufferState.js
```diff
@@ -61,6 +62,25 @@ export default class BufferState {
...
+                this.state.$off('network.connecting', onNetworkConnectingBound);
+                this.state.$off('buffer.close', onBufferCloseBound);
+            }
+        }
+
+        const onNetworkConnectingBound = onNetworkConnecting.bind(this);
+        const onBufferCloseBound = onBufferClose.bind(this);
+
+        state.$on('network.connecting', onNetworkConnectingBound);
+        state.$on('buffer.close', onBufferCloseBound);
```
prawnsalad: These 2 event handlers could do with some comments explaining why they're here, it's not obvious for someone glancing at it

PR #1191, opened 2019-10-29
Inline preview with pin-out option to the app preview panel

prawnsalad:
https://imgur.com/ejO67UC

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -447,6 +447,7 @@ body {
 .kiwi-main-mediaviewer {
     max-height: 70%;
     overflow: auto;
+    border-bottom: 1px solid rgba(0, 0, 0, 0.3);
 }
 
 .kiwi-controlinput {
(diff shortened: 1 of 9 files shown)


Shillos:
I've noticed a problem during, video chat and wanted to post a youtube video and preview it in the chat, what happened was it closed my video chat to show the youtube video and that's a problem. It would be easier if there was a way to enable inline and if anyone wants to disable to make it back to top view. There should be a switch for this and if possible to enable by default or by switch preview of videos inline.

prawnsalad:
@Shillos that would also happen on master currently. But now there is a `buffers.inline_link_previews` config option that if set to `true` it will open them inline first instead. There's a chance I might enable this by default yet.
This will however still let you pin a link and take over the conference call as there can only be one preview at the top.

Shillos:
I tried using it in the buffer in the config, but nothing helped. I still couldn't see any previews of the videos inline. Maybe just wait until you get it working properly in master. Of course, it's a problem when video chatting and you wanna show a youtube video and your video chat disappears. I will wait until this is done and I will test it again! :-) Thanks!

prawnsalad:
@Shillos does master work for you now? Inline previews are now enabled by default in master

PR #1192, opened 2019-10-31
Update IrcClient.js

dstasinski:
Added CTCP WEBSITE response to return document.referrer like Mibbit.


prawnsalad:
This would be a huge privacy leak for most people so it would be better as a plugin. https://github.com/kiwiirc/kiwiirc/wiki/Plugins

PR #1198, opened 2019-11-20
Keep BufferState in 'loading' state while processing the batch msgs

tralves:
The previous implementation of `getLoadingState` in the BufferState was returning `done` as soon as the chat history request ended. The problem is that after the messages return they will very likely be batched (see `batchedAdd.js`). As a consequence, the state `done` does not indicate that the new messages are properly loaded in the buffer, which is very important for the UI side.

This PR fixes this situation by retrieving the number of messages in the `batchedAdd.js` queue. If there are messages in this queue, keep in the loading state.


tralves:
wait. there seem to be some edge cases.

tralves:
Ok! Everything works now!

PR #1202, opened 2019-12-08
Fix Error is test starting with a number

MrGrays:
it was sending "^C1text" instead of "^C01text", which will work but if the text starts with a number the first digit of that will be interpreted as part of the colour code


ItsOnlyBinary:
this no longer appears to be an issue

PR #1213, opened 2020-01-07
Theme palettes

prawnsalad:


prawnsalad:
Once in place, we could do with adding CSS linting to these

prawnsalad:
Note for future me: This needs to be re-thought a little as in the default theme the sidebar is dark while the rest is light. In the dark theme this doesn't work as both need to be dark.

PR #1217, opened 2020-01-13
Make recaptcha a component and include it in NotConnected.vue

ItsOnlyBinary:


prawnsalad:
For completeness, I've added a quick example on how plugins can replace recptcha with their own https://github.com/kiwiirc/kiwiirc/wiki/Plugin-Examples#replace-recaptcha-with-a-custom-captcha

PR #1223, opened 2020-01-28
Make use of both avatar sizes

790:


src/components/Avatar.vue
```diff
@@ -22,13 +22,12 @@ export default {
...
         hasAvatar() {
-            return !!(this.avatar && this.avatar.small);
+            return !!(this.avatar && (this.avatar.small || this.avatar.large));
         },
         avatarStyle() {
             let style = {};
-
             if (this.hasAvatar) {
-                let url = this.avatar.small;
+                let url = (this.size === 'small' && this.avatar.small) ? this.avatar.small : this.avatar.large;
```
prawnsalad: This should probably get split up to make it easier to read at this point
~~~javascript
let url = (this.size === 'small' && this.avatar.small) ?
    this.avatar.small :
    this.avatar.large;
~~~

PR #1228, opened 2020-02-04
Safeguards repeated chathistory replies.

tralves:
**Problem:**
When the bnc replies with repeated messages in the chat history, the app fails to set the buffer `chathistory_available` tag to false, and so the app always thinks there is something to retrieve.

This is fundamentally a BNC bug, but it may be worth it to make a failsafe in the app.

**Solution:**
Checks if any message retrieved from chathistory is really new (by ID).

### DIFF
--- a/src/libs/state/BufferState.js
+++ b/src/libs/state/BufferState.js
@@ -329,9 +329,13 @@ export default class BufferState {
                     return;
                 }
 
+                // The BNC server may reply with messages that are already in the buffer.
+                // This var stores whether there are new messages in the chathistory response.
                 let hasNewMessages = event.commands.some(msg =>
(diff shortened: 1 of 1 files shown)


src/libs/state/BufferState.js
```diff
@@ -324,11 +324,15 @@ export default class BufferState {
...
             .then((event) => {
-                if (!event || event.commands.length === 0) {
+                if (!event) {
                     this.flag('chathistory_available', false);
-                } else {
-                    this.flag('chathistory_available', true);
+                    return;
                 }
+
+                let hasNewMessages = event.commands.some(msg =>
```
prawnsalad: This could do with a quick comment explaining what it's doing since it's just slightly out of context on its own here

PR #1232, opened 2020-02-12
Add user colour to left side of ControlInput.vue

ItsOnlyBinary:


prawnsalad:
While it's colourful, from a UI perspective it's really distracting. The reason nicks have different colours is to make it easier when following conversation and our eyes can pick up colours far quicker than reading letters. Adding the colour also to the nick at the bottom takes attention away from the main focus of the app which is the message list.

PR #1245, opened 2020-02-27
Translations 20200227

ItsOnlyBinary:
This replaces the other pr

closes #1110


ItsOnlyBinary:
[locales.zip](https://github.com/kiwiirc/kiwiirc/files/4262067/locales.zip)

^^^^ zip containing the converted locales. this might be importable into crowdin to help the transition but app.dev.po has been regenerated so will likely involve some manual effort for the translators. As i dont know the outcome i would try with one language first

prawnsalad:
Closing as we can now use Crowdin correctly using the existing keys and we have now learned that the i18next lib we use would break in all different ways using English as keys.

PR #1254, opened 2020-03-17
Show total of all buffers unread/highlights when network collapsed

ItsOnlyBinary:


src/components/StateBrowserNetwork.vue
```diff
@@ -202,6 +202,34 @@ export default {
         totalNetworkCount() {
             return state.networks.length;
         },
+        serverUnread() {
+            if (!this.collapsed) {
+                return this.serverBuffer.flags.unread;
+            }
+            let totalUnread = 0;
+            this.network.buffers.forEach((buffer) => {
+                if (buffer.name === '*raw') {
```
prawnsalad, line 211: This would be better checking for `buffer.isSpecial()` which covers any buffer that starts with `*`

PR #1255, opened 2020-03-20
Allow auto complete of settings when using /set

ItsOnlyBinary:


src/components/ControlInput.vue
```diff
@@ -601,6 +609,23 @@ export default {
...
 
+            if (opts.settings) {
+                let out = {};
+                let base = [];
+                settingTools.buildTree(out, base, state.getSetting('settings'), false);
+                settingTools.buildTree(out, base, state.getSetting('user_settings'), true);
+
+                let settingList = [];
+                Object.keys(out).forEach((setting) => {
+                    settingList.push({
```
prawnsalad: Any reason not to push directly to `list` here?

ItsOnlyBinary: I did it this way to match how the users/buffers are done

src/libs/settingTools.js
```diff
@@ -0,0 +1,28 @@
+'kiwi public';
+
+let ignoreKeys = ['emojis', 'themes', 'bnc', 'aliases', 'restricted', 'kiwiServer',
+    'hide_advanced', 'windowTitle', 'startupOptions', 'plugins', 'presetNetworks'];
+
+export function buildTree(data, base, object, modified) {
```
prawnsalad: This could do with a comment explaining what it does now that there is no context around it

PR #1260, opened 2020-03-24
Remove Message in buffer when traffic_as_activity is disabled

ping86:
Steps to reproduce

Use this settings:

```
"show_joinparts": false
"scrollback_size": 20, //set a low value to test more easy
```

When you join in a channel with high traffic data (join/part) more than privmsg messages this settings causes a bug. Traffic data consumes all buffer and the user only see a few messages.

We can use the setting 'traffic_as_activity' to avoid this behavior, and send to the buffer only when is necessary.

Thanks.

### DIFF
--- a/src/libs/state.js
+++ b/src/libs/state.js
@@ -539,8 +539,6 @@ const state = new Vue({
                 bufferMessage.ignore = true;
             }
 
-            buffer.addMessage(bufferMessage);
-
             // Increment the unread counter if this buffer is not active
             let includeAsActivity = false;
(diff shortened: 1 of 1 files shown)


ItsOnlyBinary:
This has been implemented with a new option `skipHiddenMessages`
if the option is `true` any new messages which would be hidden are not added to the buffer
https://github.com/kiwiirc/kiwiirc/blob/8f1ed7a50ed7dc83f492c3c1b8e680e50a272fa1/src/res/configTemplates.js#L104

PR #1267, opened 2020-04-03
Add autoReconnect setting

ItsOnlyBinary:


digitalcircuit:
Is it worthwhile exposing the other settings for auto-reconnect?

[kiwiirc/irc-framework/docs/clientapi.md](https://github.com/kiwiirc/irc-framework/blob/83937c4b16a9cd959c48cf3e81b1e267dc7ced79/docs/clientapi.md ):
```
    auto_reconnect: true,
    auto_reconnect_wait: 4000,
    auto_reconnect_max_retries: 3,
```

Those defaults seem like they'd work for me, but I imagine others may want to tweak them.

digitalcircuit:
Out of curiosity, since it's not shown in `configTemplate.js`, how would one configure the additional options, including `autoReconnect`?

My first guess [for `configTemplate.js`](https://github.com/kiwiirc/kiwiirc/blob/master/src/res/configTemplates.js ) would be…
```js
/**
 * Configuration templates
// […]
 * The advanced settings dialog also displays these settings by iterating
 * through the keys. The values must not be null otherwise they will not be
 * shown.
 */
export const configTemplates = {
    default: {
        plugins: [],
// […]
// Setting the options like this...?  Switching from camelCase to word_underscore
// ----------
        ircFramework: {
            auto_reconnect: true,
            auto_reconnect_wait: 4000,
            auto_reconnect_max_retries: 3,
        },
// ----------
        disconnectOnSaslFail: true,
        allowRegisterProtocolHandler: false,
        noticeActiveBuffer: true,
// […]
        presetNetworks: [],
    },
};
```

I'm unsure of the right balance here for discoverability and ease of configuring on the fly (the advanced settings dialog) versus not maintaining/displaying an overwhelming list of options.  Overriding `ircClient.options` directly feels like a useful feature, but perhaps not the most approachable.

Subjectively, I'd argue that auto-reconnect is a general enough feature that others may want to enable it, helping reduce the disruption of e.g. restarting `webircgateway` or network outages.

ItsOnlyBinary:
with this pr you can just create the object in your config.json and it will pass the settings to irc-fw
eg.
```
ircFramework: {
    auto_reconnect: true,
    auto_reconnect_wait: 4000,
    auto_reconnect_max_retries: 3,
},
```

if you wanted it to just be shown in advanced settings but not activated you could just set your config as...

```
ircFramework: {
    auto_reconnect: false
}
```

prawnsalad:
I'm thinking that we could add a `autoReconnect` config option that sets `ircFramework.auto_reconnect` to true?

ItsOnlyBinary:
like i did in the first commit https://github.com/kiwiirc/kiwiirc/pull/1267/commits/a0be8de3a991da1aa189752e3077e25800a08347 or a hybrid of both approaches?

prawnsalad:
Yup exactly like that. Then keeping the ircFramework options as you have for anything more specific

digitalcircuit:
This seems ideal to me, too - an easy, supported/documented setting for "I want auto-reconnect", while still allowing full customization of the underlying `ircFramework` instance.

Thank you!

PR #1268, opened 2020-04-03
Add avatar.large to UserBox.vue

ItsOnlyBinary:


AllanOcelot:
Is this going to make spacing inconsistent between users with Avatars and those without? If so, we should add placeholders for users without.

ItsOnlyBinary:
superseded by #1547

PR #1270, opened 2020-04-07
[API] MediaViewer fixes and improvements

ItsOnlyBinary:
this pr achieves a few things:

- Allows props to be passed with a component

- Allows plugins to better control the mediaviewer height (by not having component/iframe in a extra div)

- Stops the loading of embedly script when a component/iframe is used

- Fixes switching from embedly to a component/iframe without closing the mediaviewer

- Updates MediaViewer.vue to more modern code style


ItsOnlyBinary:
rebased

PR #1279, opened 2020-04-14
Override emojis when loading from config

ItsOnlyBinary:
see #1284 for an alternative

closes #1284


prawnsalad:
Closing in favour of #1284

PR #1284, opened 2020-04-21
Overwrite config objects when key ends with _overwrite

ItsOnlyBinary:
more generic alternative solution for #1279

closes #1279


prawnsalad:
I'm not sure this is the cleanest way to go about this. How about having a list of full key names that gets entirely replaced instead? eg `let mergeObjects = ['emojis', 'some.key'];`

ItsOnlyBinary:
Superseded by #1566

PR #1291, opened 2020-04-26
Add StateBrowser headers

CoryChaplin:
The statebrowser today is a simple list of channels and queries.
To make it more readable for the user and to allow improved focus, this is a proposal to add section titles that can fold up to gain some vertical space in the statebrowser.
This is what it looks like.

header


src/components/StateBrowserNetwork.vue
```diff
@@ -438,11 +503,24 @@ export default {
...
+    font-size: 0.8em;
+    text-transform: uppercase;
+    margin-left: 5px;
+}
+
+.kiwi-statebrowser-buffers-toggle {
+    width: 10px;
 }
 
-.kiwi-statebrowser-channels-option {
```
prawnsalad, line 445: This change is technically correct and this class was badly named originally. But this will need all themes updating and would break any third party themes too - a much bigger task so we've left this alone for now.

src/components/StateBrowserNetwork.vue
```diff
@@ -355,6 +410,17 @@ export default {
             this.channel_filter_display = !this.channel_filter_display;
             this.channel_add_display = false;
         },
+        toggleSection(type) {
+            if (type === 'channels') {
+                this.section_display_channels ?
```
prawnsalad, line 415: `this.section_display_channels = !this.section_display_channels;` :)

prawnsalad:
Something this would need, is the green/red label counting the highlights and unread messages after "Channels" and "Messages" when they are collapsed.

Not sure if it does it already, but also the Messages section should be removed if there are no query buffers available.

src/components/StateBrowserNetwork.vue
CoryChaplin: OK rolled that back.

src/components/StateBrowserNetwork.vue
CoryChaplin: Thanks, fixed that.

CoryChaplin:
> Something this would need, is the green/red label counting the highlights and unread messages after "Channels" and "Messages" when they are collapsed.
> 
> Not sure if it does it already, but also the Messages section should be removed if there are no query buffers available.

Implemented this.

src/components/StateBrowserNetwork.vue
```diff
@@ -105,61 +105,113 @@
...
+                            <div class="kiwi-statebrowser-buffer-actions">
+                                <div class="kiwi-statebrowser-channel-labels">
+                                    <div
+                                        v-if="(!section_display_channels === true &&
+                                            type === 'channels') ||
+                                            (!section_display_queries === true &&
+                                                type === 'queries')"
+                                        class="kiwi-statebrowser-channel-label"
+                                    >
+                                        {{ buffersUnread(itemBuffers) }}
```
CoryChaplin, line 163: I'm unsure how to manage the 0 and the highlights here without adding multiple calls to buffersUnread().
@prawnsalad is it better to move buffersUnread() to the computed section ? If so, I don't see how to catch the filteredBuffers.

CoryChaplin: Discussed this on IRC.

CoryChaplin:
@prawnsalad done the improvements we spoke about here.

gokturkyilmaz:
That feature is so complicated for users and not useful. It should be more easier to use. 👎

prawnsalad:
@CoryChaplin hmmm look like I messed this up in my late night refactoring. Will tidy it back up.

@gokturkyilmaz All this does is add a header for channels and queries, nothing changes for usability

gokturkyilmaz:
@prawnsalad, does it (**CHANNELS** and **MESSAGES** tags) will also shown in available other languages?

westor7:
That PR is interesting, a small suggestion is if it is possible for an configuration option to be able to enable/disable it.

prawnsalad:
@gokturkyilmaz yes. 

@westor7 im trying not to as it's extra maintenance burden. Any particular reason you would want it disabled?

CoryChaplin:
@westor7 it's not so much a tree than adding a header to the existing thing above channels and addind a separator to identify where queries start.
It also opens the possibility to add better ways to create/search channels and PM's (inspired by Slack & co).

CoryChaplin:
@prawnsalad tested on all themes, everything works smoothly in my eye

westor7:
> 
> 
> @westor7 it's not so much a tree than adding a header to the existing thing above channels and addind a separator to identify where queries start.
> It also opens the possibility to add better ways to create/search channels and PM's (inspired by Slack & co).

I'm not negative with this mate, its seems making things more easy (at least for my eyes), i was talking for these people that doesn't want that.

prawnsalad (approved):

PR #1299, opened 2020-05-03
PASSWORD SERVER URL

elvis72500:
Hello, I would like to know if we can add to transmit in the url the server password?
Because it allows me to identify the nickname directly via the server.
cordially


prawnsalad:
A password in the URL is a very bad idea from a security standpoint as it will be included in browser history and server logs. If you want to include the password in your page then you can write your own kiwi plugin to read it from somewhere (like the URL) and set it yourself

PR #1301, opened 2020-05-04
Fix inline layout by targeting .kiwi-messagelist-message--inline

ItsOnlyBinary:


ItsOnlyBinary:
Did not work. Needs revisiting

PR #1305, opened 2020-05-06
Allow setting recaptcha api url via config

ItsOnlyBinary:
This allows the use of the hcaptcha api instead of googles recaptcha

i have done my best to write some documentation here:
https://gist.github.com/ItsOnlyBinary/0318b00bd831f57ede7f8bd5e210ff5c

### DIFF
--- a/src/components/Captcha.vue
+++ b/src/components/Captcha.vue
@@ -12,6 +12,7 @@ export default {
     props: ['network'],
     data() {
         return {
+            recaptchaURL: '',
             recaptchaSiteId: '',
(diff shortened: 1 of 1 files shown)


src/components/Captcha.vue
```diff
@@ -12,6 +12,7 @@ export default {
     props: ['network'],
     data() {
         return {
+            recaptchaURL: '',
```
prawnsalad, line 15: `recaptchaURL` should be `recaptchaUrl`. Acronyms should always be in camel case if over 2 chars long

PR #1316, opened 2020-05-11
disable parent window scroll when kiwiirc used inside iframe

mohan43u:
when kiwiirc used inside iframe, the parent window jumps into kiwiirc iframe everytime there is new message in kiwiirc. This patch disables jumping.

### DIFF
--- a/src/components/MessageList.vue
+++ b/src/components/MessageList.vue
@@ -450,7 +450,7 @@ export default {
                 return;
             }
             if (this.$state.setting('enable_parent_scroll')) {
-                    this.$refs.ender.scrollIntoView(false);
+                this.$refs.ender.scrollIntoView(false);
             }
         },
         maybeScrollToBottom() {
--- a/src/components/MessageList.vue
+++ b/src/components/MessageList.vue
@@ -449,7 +449,9 @@ export default {
                 this.$el.scrollTop = this.$el.scrollHeight;
                 return;
             }
-            this.$refs.ender.scrollIntoView(false);
+            if (this.$state.setting('enable_parent_scroll')) {
+                    this.$refs.ender.scrollIntoView(false);
+            }
         },
         maybeScrollToBottom() {
             if (!this.maybeScrollToBottom_throttled) {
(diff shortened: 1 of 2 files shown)


ItsOnlyBinary:
This would completely disable scrolling for new messages.

It would be better to create a computed `isIframe()` like https://github.com/kiwiirc/kiwiirc/blob/master/src/components/MessageList.vue#L233 and find a way to auto detect if its an iframe reliably.

Then change https://github.com/kiwiirc/kiwiirc/blob/master/src/components/MessageList.vue#L446 to `if (this.isIos || this.iframe)`

mohan43u:
> This would completely disable scrolling for new messages.
> 
> It would be better to create a computed `isIframe()` like https://github.com/kiwiirc/kiwiirc/blob/master/src/components/MessageList.vue#L233 and find a way to auto detect if its an iframe reliably.
> 
> Then change https://github.com/kiwiirc/kiwiirc/blob/master/src/components/MessageList.vue#L446 to `if (this.isIos || this.iframe)`

Reworked as per suggestion

ItsOnlyBinary:
nice work mohan34u :)

two last things to make it perfect. the config option should no longer be needed, this is something that will affect all kiwi's in iframes

also `return window !== window.parent;` is a little clearer to read

mohan43u:
is it acceptable if we omit the current functionality? like what if someone want to have a jump to iframe if something happens in IRC?

perhaps we can make 'disable_parent_scroll: true" in configTemplates.js

ItsOnlyBinary:
From a user perspective it would be very annoying if you was trying to read a page and it jumped because of a message in a component of that page.

So i cannot see why anyone would want that option. But not too much trouble if prawn disagrees with me to readd the options back.

mohan43u:
> From a user perspective it would be very annoying if you was trying to read a page and it jumped because of a message in a component of that page.
> 
> So i cannot see why anyone would want that option. But not too much trouble if prawn disagrees with me to readd the options back.

removed 'disable_parent_scroll' config option

mohan43u:
any chance of fixing this issue?

PR #1322, opened 2020-05-13
Make message senders nick double clickable

### LINKED ISSUE #1321: quoting a user name by double-clicking
When I double click on user name on their "join" action, their nick is copied into message line.

But that feature is not working when that user sends a message (it works only on room "join" action lines). Single click shows that user info, but double-click would be more helpful for easier and quicker quote-reply.

ItsOnlyBinary:
This makes sender nicks linked with click and double click support and improves double click nick insert to have matching case to the users nick

also fixes click on realname in modern layout (missing `prop.`)

closes #1321

### DIFF
--- a/src/components/MessageList.vue
+++ b/src/components/MessageList.vue
@@ -381,9 +381,11 @@ export default {
 
             let network = this.buffer.getNetwork();
             let user = network.userByName(dataNick);
-            if (user && user.nick) {
-                this.$state.$emit('input.insertnick', user.nick);
-            }
+            // The user might have left use dataNick as fallback
+            let nick = (user && user.nick) ?
+                user.nick :
+                dataNick;
+            this.$state.$emit('input.insertnick', nick);
         },
         onMessageClick(event, message, delay) {
             // Delaying the click for 200ms allows us to check for a second click. ie. double click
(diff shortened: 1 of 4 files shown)


src/components/MessageListMessageModern.vue
```diff
@@ -161,9 +161,9 @@ const methods = {
         }
 
         // No point showing the realname if it's the same as the nick
-        if (props.message.user.nick.toLowerCase() === props.message.user.realname.toLowerCase()) {
-            return false;
-        }
+        // if (props.message.user.nick.toLowerCase() === props.message.user.realname.toLowerCase()) {
```
prawnsalad, line 164: Commented lines left over from debugging I'm guessing?

src/components/MessageList.vue
```diff
@@ -374,9 +374,15 @@ export default {
...
-            if (userNick) {
-                this.$state.$emit('input.insertnick', userNick);
+            let dataNick = event.target.getAttribute('data-nick');
+            if (!dataNick) {
+                return;
+            }
+
+            let network = this.buffer.getNetwork();
+            let user = network.userByName(dataNick);
+            if (user && user.nick) {
```
prawnsalad, line 384: This will block it from inserting the nick if the user has already left the channel. It should use `dataNick` if the user object doesn't exist

CoryChaplin:
Tested this and it works as it should 👍

PR #1331, opened 2020-05-21
Restore bnc username/password from bnc network.

tralves:
**Problem**
When the app is using a bouncer, the bouncer sets its credentials when connecting to the bnc network on an `irc.motd` listener. The problem happens when the app loads its networks from the StatePersistence, the bnc credentials will not be set when the app connects to a (non-bnc) network.

**Solution**
This PR makes sure the `BouncerProvider.js` loads the username and password from the BNC network connection on `enable()`.

It also makes sure the network attributes `hidden` and `is_bnc` are persisted.

### DIFF
--- a/src/libs/BouncerProvider.js
+++ b/src/libs/BouncerProvider.js
@@ -47,6 +47,16 @@ export default class BouncerProvider {
         this.bnc.path = path || '';
         this.bnc.enabled = true;
 
+        const bncNetwork = this.state.networks.find((network) => network.is_bnc);
+
+        if (bncNetwork?.connection?.password) {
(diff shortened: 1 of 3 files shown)


src/libs/BouncerProvider.js
```diff
@@ -47,6 +47,14 @@ export default class BouncerProvider {
         this.bnc.path = path || '';
         this.bnc.enabled = true;
 
+        const bncNetwork = this.state.networks.find((network) => network.is_bnc);
```
prawnsalad, line 50: This could do with a quick comment explaining why it's looking for a is_bnc network and why it's reading the password field

PR #1341, opened 2020-05-29
Avatar in UserBox

H7-25:
Added avatar in UserBox


ItsOnlyBinary:
Due to this pr being broken i've redone this feature but with a bigger avatar see #1547

PR #1350, opened 2020-06-05
Encourage users to change the default nick; Also adds some more events for InputText.vue

ItsOnlyBinary:


prawnsalad:
What problem is this solving?

It looks to be simulating placeholder text in a non-standard way?

ItsOnlyBinary:
I wanted to select the default nick to encourage users to change it but that looked ugly. so instead just typing with the nick input selected and the cursor at the end will remove the default nick.

If escape is pressed or focus is lost when the nick box is empty the default nick will be reapplied

PR #1356, opened 2020-06-10
Send PART before emitting buffer.close event

ItsOnlyBinary:


half-duplex:
I assume that was done so that plugins could do things before PART, and any that do would be broken by this change. Why do you want PART first?

prawnsalad:
@ItsOnlyBinary ?

ItsOnlyBinary:
This change is to play nice with kiwibnc. the part cannot be prevented and was causing kiwibnc to reopen the buffer to display error about parting a channel you are no longer in. the buffer object still exists when `buffer.close` is emitted so should not break any plugins

half-duplex:
I wasn't referring to plugins expecting the buffer to exist, but plugins assuming they can still e.g. PRIVMSG a +n channel, which would be broken by this change.

Could you explain the order of events with kiwibnc that ends with the buffer re-opening?

PR #1357, opened 2020-06-13
Add ability to set custom username on websockets

### LINKED ISSUE #1345: Configure username, realname and send_quit_on_client_close when using direct websocket connection
When connecting directly to IRCd using websockets (without webircgateway), it would be nice to be able to configure the following in config.json:
- username (mentioned in the source code as network.username)
- realname (mentioned in the source code as network.gecos)
- send_quit_on_client_close

Those options are present in webircgateway [configuration file](https://github.com/kiwiirc/webircgateway/blob/master/config.conf.example).

westor7:
You can use "username": "whatever", in config.json file to set your custom ident that you want the websockets clients to use.

Closes #1345

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -290,6 +290,7 @@ export default {
                 direct: connectOptions.direct,
                 path: connectOptions.direct_path || '',
                 gecos: options.gecos,
+                username: options.username,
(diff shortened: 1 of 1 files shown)


westor7:
Credits on @ItsOnlyBinary for the code.

PR #1358, opened 2020-06-14
Add option for a custom quit message on websockets

westor7:
This allows you to set ```quitMessage``` option on config.json if you use ```startupOptions.direct``` option, when ever the browser or tab will be closed the websocket will send now a QUIT with a message (like on webircgateway via webirc).
If not any ```quitMessage``` option is set in config.json the default quit will be ```Client Closed Connection```

Config.json Example:

```"quitMessage": "My websocket connection has been closed",```

Credits on @ItsOnlyBinary for the code.


westor7:
Also requested before on #1345

PR #1367, opened 2020-06-19
Don't try to send typing TAGMSG's to buffers that wont accept them

### LINKED ISSUE #1366: Don't send tag messages to none joined/moderated channels
Banned from channel but while typing a command got `You cannot send external messages to this channel whilst the +n (noextmsg) mode is set.`

@westor7 has also reported that +m channels are also an issue for users without >= +v

ItsOnlyBinary:
closes  #1366

### DIFF
--- a/src/libs/state/BufferState.js
+++ b/src/libs/state/BufferState.js
@@ -412,6 +412,27 @@ export default class BufferState {
         );
     }
 
+    hasMode(mode) {
+        return Object.keys(this.modes).indexOf(mode) > -1;
+    }
+
+    shouldShareTyping() {
+        let network = this.getNetwork();
+        if (!this.setting('share_typing')) {
+            return false;
+        }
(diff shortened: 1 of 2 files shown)


ItsOnlyBinary:
I do wonder if the startTyping / stopTyping logic in ControlInput.vue should be completely moved to TypingMiddleware.js so its all together.

src/libs/state/BufferState.js
```diff
@@ -412,6 +412,27 @@ export default class BufferState {
...
+        if (!this.setting('share_typing')) {
+            return false;
+        }
+        if (!this.isChannel() && !this.isQuery()) {
+            return false;
+        }
+        if (this.isChannel() && !this.joined) {
+            return false;
+        }
+        if (this.hasMode('m') && !this.userMode(network.currentUser())) {
```
prawnsalad, line 430: This could do with a quick comment explaining what exactly it's checking for

PR #1368, opened 2020-06-20
Sychronize some hex colors with var()

westor7:
I think it's better now because it is syncing with the --brand-* colors.


prawnsalad:
Closing this PR as we have https://github.com/kiwiirc/kiwiirc/pull/1213 which will replace all the current vars anyhow

PR #1369, opened 2020-06-21
Fix welcome page not ready if channel is invalid

### LINKED ISSUE #1310: No hashtag in front of Channels leads to a private chat
I don't know if this is a bug or something missing, but when I use this without the # in front of a channel it leads me after connecting, to a nick that doesn't exist. Although it should have been a channel instead.
No hashtag on channel 1
No hashtag on channel 2
I hope this can be resolved!

westor7:
It checks if the first character is # that determines that is a channel, if it is not it doesn't allowing to connect.

Closes #1310

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -143,6 +143,11 @@ export default {
             if (!this.connectWithoutChannel && !this.channel) {
                 ready = false;
             }
+            
+            // Checking if in #channel the first character is #
+            if (!this.connectWithoutChannel && this.channel && this.channel[0] !== "#" {
+                ready = false;
+            }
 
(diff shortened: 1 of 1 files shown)


Shillos:
Oh, finally some fixes on that #1310 I was wondering about it just today. :-)

Shillos:
Hmm, this is not entirely right though! when someone types in the channel area without the hashtag it should also allow the user to still join the channel, like in TheLunge Chat they have the same way even without the # it still joins a channel. Though I thought because Kiwi is written the same way with IRC-framework it should do the same.

Shillos:
Of course, it will do for now, until there is another way round! :-)

PR #1374, opened 2020-06-29
Add invite user to channel in query sidebar

### LINKED ISSUE #1372: Add a 'Invite user to..' option in the query sidebar
It should list all of our channels

ItsOnlyBinary:
It could maybe look nicer but I'm unsure how to achieve that

closes #1372

### DIFF
--- a/src/components/SidebarAboutBuffer.vue
+++ b/src/components/SidebarAboutBuffer.vue
@@ -168,9 +168,7 @@ export default {
             if (!this.inviteNick) {
                 return;
             }
-            // TODO switch these back once irc-framework is updated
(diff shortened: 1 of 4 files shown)


CoryChaplin:
My feeling is that the UI would be better if the invitation was possible from the target channel, within a UI where you can search users that are in common channels. Maybe a complement to this.

prawnsalad:
`/late` @CoryChaplin yup that would make sense. @ItsOnlyBinary I guess the same type of thing in the channel sidebar but lists other users from the network? I can take over with the styling if needed once this part is in

ItsOnlyBinary:
rebased

PR #1381, opened 2020-07-03
Welcome: Use saved password

half-duplex:
Kiwi already stores the network password in localStorage, let's not make the user retype it each time. This is particularly useful for embedded clients with a nontechnical audience, either on servers with a password or servers where the server password is automatically forwarded to NickServ.

Whether it's a good idea to store it at all without the user's explicit request is debatable but a different issue.

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -214,9 +214,17 @@ export default {
         } else {
             this.nick = options.nick;
         }
-
         this.nick = this.processNickRandomNumber(this.nick || '');
-        this.password = options.password || '';
+
+        if (options.password) {
+            this.password = options.password;
+        } else if (previousNet && previousNet.password) {
+            this.password = previousNet.password;
+            this.show_password_box = true;
+        } else {
+            this.password = '';
+        }
+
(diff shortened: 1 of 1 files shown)


prawnsalad:
Nice, yea this would be handy. It should use the password from the options first as it currently does though, and then use the remembered password if that's empty.

half-duplex:
Hmm. Is there ever a reason for a user to use a password other than the options one? If not, that makes sense. If there is, maybe we could save the password only when it was modified? Though that would be confusing if it were accidentally modified...

prawnsalad:
If I connect to a network using options.password, it then gets remembered. If the password changes and updated in options.password, it would never use it again as it would continue using the remembered one.

As an admin configuring my webirc, the only reason I would pre-set a password in options.password is if I want that password to be used. The user could change the password in the UI if they wanted for some reason, but I'm not sure why you would want to given the webirc is telling you to use a pre-set.

half-duplex:
Updated to config > saved > empty

prawnsalad (approved):

prawnsalad:
Sorry this took so long to review again, thanks!

PR #1382, opened 2020-07-03
Welcome: Query param to hide channel field

half-duplex:
This adds the ability to hide the channel field from the Welcome page using `?showChannel=false`. The channel field isn't hidden if no channel is provided in the config or the URL.
This is useful for embedded clients that are only meant for use with one channel, reducing user confusion and simplifying the connection form.

Wiki page to update on merge: https://github.com/kiwiirc/kiwiirc/wiki/Linking-to-and-Embedding-Kiwi-IRC#options

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -218,9 +218,13 @@ export default {
         this.nick = this.processNickRandomNumber(this.nick || '');
         this.password = options.password || '';
         this.channel = decodeURIComponent(window.location.hash) || options.channel || '';
-        this.showChannel = typeof options.showChannel === 'boolean' ?
-            options.showChannel :
-            true;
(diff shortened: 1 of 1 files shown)


prawnsalad:
This option is available in config.json already via `startupOptions:{showChannel:true}`. We don't add all these options to the URL since that can get very messy.

For websites embedding kiwi, they can customise the config dynamically using one of these options https://github.com/kiwiirc/kiwiirc/wiki/Configuration#use-a-javascript-function

sergip76:
I use Kiwi IRC embedded in an iFrame. How can I use a Javascript function to change start options?
Is there any example?

prawnsalad:
@sergip76 the javascript config function is only available if you host kiwi yourself. If you're using kiwiirc.com then you can take a look at https://kiwiirc.com/clientbuilder/ which provides more config options

PR #1399, opened 2020-07-23
Send 'notification.enabled' event

ItsOnlyBinary:
This allows plugins to use notifications without having to handle requesting permissions


prawnsalad:
In what case would `notification.enabled` be used in plugins? Plugins shouldn't need to request notification access as that's handled already and only at a certain point for which browsers trust the event, he first click after connected to a network

PR #1436, opened 2020-11-18
Make bouncer provider better at syncing networks

ItsOnlyBinary:


prawnsalad:
This looks more stable now, nice.

https://github.com/kiwiirc/kiwiirc/blob/5836062bfd7633f624ff4a7213c3dd03539f82f9/src/libs/BouncerProvider.js#L462 is there any assurance that this event will be be triggered? I can't see how it's expected to be called reading through this source. Would a lock around `saveState` be handy so that you can trigger saveState() after adding the event which ensures its only run once at a time?

ItsOnlyBinary:
That event is not needed after its been synced, and sync is started when the details change. its only there to catch the people who click connect before the debounce has timed out

PR #1445, opened 2020-12-08
Fix a layout problem on mobile browsers when selfuser vue is open

H7-25:
fix a layout problem on mobile browsers when in portrait mode and selfuser vue is open

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -895,6 +895,11 @@ export default {
     .kiwi-controlinput-portrait-hide {
         display: none;
     }
+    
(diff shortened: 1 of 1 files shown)


prawnsalad:
Do you have a before/after screenshot of what this fixes? I'm not exactly sure why hiding the controlinput would change anything

H7-25:
After and Before: 
Simosnap IRC Network - Kiwi IRC Client
Simosnap IRC Network - Kiwi IRC Client- 2

prawnsalad (approved):

PR #1449, opened 2020-12-14
Allow disabling url auto previews in queries

ItsOnlyBinary:


prawnsalad:
What's the case for wanting this?

ItsOnlyBinary:
`[14 17:06:48] Overnumerousness I have file uploader generating a embedly card for all links in chat. Can I disable this feature in private messages since there's potential they don't want those links crawled by a 3rd party service?`

seems logical to maybe not want the auto embed for messages that are private

PR #1450, opened 2020-12-18
Add support for fileuploader's new embed feature

ItsOnlyBinary:


prawnsalad:
The `addEventListener('message', this.messageEventHandler)` looks a bit awkward hiding behind the flag. Can it not be added in `created()` and removed in `beforeDestroy()` without needing flags? There is also the `this.$listen(window, 'message', fn)` function you can call in `created()` and that handles the removal of the event automatically.

src/components/MediaViewer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="kiwi-mediaviewer">
+    <div ref="mediaviewer" class="kiwi-mediaviewer">
```
prawnsalad, line 2: This ref can already be referenced via `this.$el`

prawnsalad:
While in here it's probably also wise to add some sandbox options to each iframe, just so that any embedded content can't redirect the kiwi page away.

As a computed prop:
~~~
        iframeSandboxOptions() {
            // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox
            // Mostly all permissions other than allow-top-navigation so that embedded content
            // cannot redirect the page away from kiwi
            let options = [
                'allow-downloads',
                'allow-forms',
                'allow-modals',
                'allow-orientation-lock',
                'allow-pointer-lock',
                'allow-popups',
                'allow-popups-to-escape-sandbox',
                'allow-presentation',
                'allow-same-origin',
                'allow-scripts',
            ];

            return options.join(' ');
        },
~~~

And added `:sandbox="iframeSandboxOptions"` to each iframe

ItsOnlyBinary:
superseded by #1469

PR #1472, opened 2021-01-11
Fix input wrapping

### LINKED ISSUE #1405: Input Box larger then text field
**Describe the bug**
The visual box for Text input is larger than the input form field. So you can click in the box without setting focus to the text input. If clicked on the wrong spot, instead of Text hotkeys are accepted, producing unwanted behaviour

**To Reproduce**
1. open any given kiwichat instance (e.g. freenode)
2. Move mouse pointer to the upper boundary of the input field
3. Mouse pointer changes from cursor to pointer (see screenshot) while still inside the box
4. click

**Expected behavior**
Focus the input field

**Screenshots**
kiwiirc_input

 - OS: [e.g. Windows 10]
 - Browser Chrome
 - Version 84 and before
 - Also seen on Firefox

CoryChaplin:
This makes the input box easier to select by giving it a wider zone without changing appearance.

closes #1405

### DIFF
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -781,7 +781,6 @@ export default {
     height: 100%;
     box-sizing: border-box;
     overflow: visible;
-    padding: 7px 0 12px 0;
 }
 
 .kiwi-controlinput-tool {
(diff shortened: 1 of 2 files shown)


AllanOcelot (approved):
This looks like a good solution to the problem.

PR #1473, opened 2021-01-11
Make captcha always show when needed

### LINKED ISSUE #1470: Move captcha to a modal dialog
**Describe the bug**
If the recaptcha is ever removed from the DOM (ie. switching buffers) then it breaks. Depending on the UI state if you're in the network settings, a buffer, app settings, the recaptcha UI may not be added to the UI to complete when a network connects.

**Expected behavior**
The recaptcha component should be available in the UI at any point in the app lifecycle.

It could be as a model dialog that takes over the entire screen, or a notification tab that appears somewhere on the screen waiting for you to open it. Model dialog is most likely easier.

ItsOnlyBinary:
closes #1470

### DIFF
--- a/src/components/Captcha.vue
+++ b/src/components/Captcha.vue
@@ -1,6 +1,6 @@
 <template>
-    <div>
-        <div v-if="showCaptcha" ref="captchacontainer" />
+    <div class="kiwi-captcha-outer" :class="{'kiwi-captcha-show': visible}">
+        <div ref="container" class="kiwi-captcha-container" />
     </div>
 </template>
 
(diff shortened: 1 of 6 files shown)


src/components/Captcha.vue
```diff
@@ -9,68 +9,109 @@
...
         },
-        recaptchaSuccess(response) {
-            this.recaptchaResponse = response;
+        captchaSuccess(response) {
+            this.response = response;
 
-            // If we have a network instance already, send the captcha response
-            if (this.network && this.network.state === 'connecting') {
-                this.network.ircClient.raw('CAPTCHA', response);
+            this.networks.forEach((network) => {
```
prawnsalad, line 71: This part seems a little weird. Each network captcha request would have a difference response wouldn't they? So sending the same response to every network would cause them to fail?

ItsOnlyBinary: WIG only uses one captcha key, so it should be fine for multiple networks to validate the same response,

i shall try to find a way to test it

ItsOnlyBinary: then again i might be totally wrong :(

need to look into this some more

PR #1475, opened 2021-01-15
Update +typing to the final spec version

hhirtz:
Now that the specification has been ratified, kiwiirc should use
`+typing` to interoperate with other clients.

Link to the spec: <https://ircv3.net/specs/client-tags/typing>


linuxomaniac (commented):
lgtm

prawnsalad (approved):

PR #1476, opened 2021-01-20
Allow channel name from location query params

Eluminae:
We cannot use the location anchor to prefill the channel with a direct
private channel to some user nick. As "#foobar" will previll the public
channel "#foobar" and not "foobar"

This patch allow the query parameter "?channel=foobar" to prefill the
channel input with a direct private channel and not a public one.

This will not break existing links cause we do not change the previous
url anchor way.

### DIFF
--- a/src/components/startups/Welcome.vue
+++ b/src/components/startups/Welcome.vue
@@ -225,7 +225,8 @@ export default {
             this.password = '';
         }
 
-        this.channel = decodeURIComponent(window.location.hash) || options.channel || '';
+        const params = new URLSearchParams(window.location.search);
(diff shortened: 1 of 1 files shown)


ItsOnlyBinary:
Unfortunately I'm not going to merge this change due to the possibility of confusion in using it.

`?channel=#channel` although would work, query param `channel` would be empty.

My suggested solution for your use case would be to generate a per user config, that includes the target nick in `startupOptions.channel` and set `startupOptions.allowNoChannel` to `true`,

see here for ways to include a per user config https://github.com/kiwiirc/kiwiirc/wiki/Configuration

PR #1479, opened 2021-01-22
Fix autocomplete not completing and emptying input in particular contexts.

### LINKED ISSUE #1388: Nickname autocompletation after a command don't work
Nickname autocompletation no more work if the nick is after a command

/kick nickname (press tab to autocomplete the nickname)

### LINKED ISSUE #1471: /me blocks completing the nickname via tab
Tabbing for completing the nickname/channel does not use if using **/me** infront

CoryChaplin:
This fixes 2 bugs:

- When autocomplete has already been triggered, it wont be triggered again. This breaks usages like /kick nick+tab. If the user has typed a space or a semi-colon, the auto-complete should be reset.
- When typing something+tab+continue typing and the tab doesn't match any auto-complete proposal, the input is reset and it should not.

closes #1388, closes #1471

### DIFF
--- a/src/components/AutoComplete.vue
+++ b/src/components/AutoComplete.vue
@@ -143,6 +143,8 @@ export default {
 
             let cancelKeyCodes = [
                 13, // return
+                32, // space
+                186, // semi-colon
                 188, // comma
                 190, // period
             ];
--- a/src/components/ControlInput.vue
+++ b/src/components/ControlInput.vue
@@ -372,7 +372,7 @@ export default {
         },
         onAutocompleteSelected(selectedValue, selectedItem) {
             let word = selectedValue;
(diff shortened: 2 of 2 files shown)


prawnsalad:
Could you paste a quick gif of these 2 things before+after? I can't seem to replicate these issues on kiwiirc.com

CoryChaplin:
1. Tab on second word
**Before:**

tab

**After:**
tab-ok

2. Tab with no matches
**Before:**
lost

**After:**
lost-ok

Curiously, the first issue doesn't appear on Kiwiirc.com but it does on master. Maybe a local fix or a recent regression ?

DarkTexas:
is there any update on this?

H7-25:
Very wanted fix ;)

PR #1483, opened 2021-01-31
FIX: Navcontainer active items overlap by one pixel when active

AllanOcelot:
A super quick fix that reduces the line height of some child items to ensure that the navContainer items do not exceed the parent.

Without fix:
image

With Fix:
image


prawnsalad (approved):

prawnsalad:
Nice, that has annoyed me

Issue #1231, opened 2020-02-11
Big nicknames doesn't displaying correctly in nicklist


westor7:
Hello,

I saw that when a user uses a big nickname about 25+ nick chars then kiwiirc doesn't displaying correctly that nickname in nicklist.

I tried to use `.kiwi-nicklist-user-nick { text-overflow: ellipsis; }` but didn't fixed.

Image Example: https://i.imgur.com/fT7GyQc.png

- Thanks!

ItsOnlyBinary:
This should now be fixed in master

PR #1484, opened 2021-02-01
Support for longer nicknames in the Users Nicklist sidebar

AllanOcelot:
- Fix for #1231 
- Style change to the nicklist name, ensuring long names have ellipses
- Add padding to the right side of nickname so the 'message' icon is never obscured 

Without Fix:
image

With Fix:
image


ItsOnlyBinary:
closes #1231

PR #1490, opened 2021-02-10
Allow setting of embedly script url in config.json

ItsOnlyBinary:


prawnsalad (approved):

prawnsalad:
> itsonlybinary: that url is purely for the embedly webpreview, i added it because in electron the url was resolving // to file://, so i needed to force it https://

PR #1504, opened 2021-03-02
Add topic in header with irc markup

CWO-Stanley:
This adds the topic in the header including the colors as it is shown on IRC itself.
Topic is same as the channelname, ellipsis when too long.
when hovering channelname, channelname gives a background for better reading as the :hover gives z-index 1

### DIFF
--- a/src/components/ContainerHeader.vue
+++ b/src/components/ContainerHeader.vue
@@ -8,6 +8,9 @@
             <div class="kiwi-header-name-container">
                 <div class="kiwi-header-name">{{ buffer.name }}</div>
             </div>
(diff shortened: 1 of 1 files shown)


CoryChaplin:
@CWO-Stanley It's maybe worth sharing a couple of before/after screenshots.

CWO-Stanley:
81c1f019-96d8-464a-acaf-8d9a53dba666

CoryChaplin:
I think the design could be improved to have a better hierarchy of information.

For example Slack does it this way: 
Slack___alerting-atome___Bedrock

And Discord does it this way:
general_-_Discord

PR #1507, opened 2021-03-06
Fix a style issue on "Join Channel" button

westor7:
The button "Join Channel" was not clearly displayed, tested on all themes and this change seems to fix that issue.

Before:
before

After:
after


ItsOnlyBinary:
I feel this is something that should just be fixed in the nightswatch theme. The other themes appear to work as intended

PR #1516, opened 2021-04-03
Update app.sv-SE.po

Umeaboy:
Signed-off-by: Kristoffer Grundström <lovaren@gmail.com>

### DIFF
--- a/src/res/locales/app.sv-SE.po
+++ b/src/res/locales/app.sv-SE.po
@@ -21,67 +21,67 @@ msgstr "Start"
 
 #: Welcome to Kiwi IRC!
 msgid "start_greeting"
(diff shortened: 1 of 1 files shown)


prawnsalad:
Hi, thank you for translating this! We do use a translation manager at https://translations.kiwiirc.com/ for these files so that everybody involved can be in sync. It also lets us know just how much has been translated and what can be imported into the project.

It's awesome to see Swedish included as I have seen several Swedish communities using Kiwi IRC, but lets make sure we can incorporate it correctly!

Umeaboy:
Can I safely delete the branch now then?

prawnsalad:
@Umeaboy yes if the changes have been ported over then its safe to remove the branch!

Issue #1146, opened 2019-08-26
[Suggestion] TAGMSG: Make the dots animation!


gokturkyilmaz:
Is it possible to make the dots _(**…**)_ animations, when someone(s) typing a message in PM or channel like Skype, facebook….

AllanOcelot:
Yes, this could easily be done by someone using CSS animations.
Would be a great first commit for someone to the project I think :)

Shillos:
Here is something that might give you an idea https://codepen.io/rajarju/pen/pFrIt

westor7:
@Shillos nice idea about https://nzbin.github.io/three-dots/ , looking very pretty..

PR #1518, opened 2021-04-14
Animated typing indicator

CWO-Stanley:
this adds a animated typing indicator when users are typing instead of the standing 3 dots

#1146


CWO-Stanley:
000a2c84-5986-485c-ac03-bfeb8b395538
24d5a761-2f66-49f4-9b4a-32ea5692e294
d51e80fa-9510-4bdf-977c-2017dceca3d0


Sorry for not having animated screens :(

PR #1523, opened 2021-04-16
Statebrowser groups multiple fixes/improvements

ItsOnlyBinary:
- Fix showing buffers other than channel/query eg `*raw` or `*bnc`
- Don't show messages group when there are no queries
- Add setting to enable/disable buffer groups


ItsOnlyBinary:
@CoryChaplin maybe you want to review these changes ;)

CoryChaplin (approved):
Looks good to me.
For the log, maybe useful to give some examples of buffers that are not channels or queries ?

PR #1527, opened 2021-04-19
Fix channel list layout on narrow screens

CoryChaplin:
### Before
before

### After
after


H7-25:
I've choice another solution, just using additional css, topic keep to much space
Schermata 2021-04-19 alle 23 40 50

CoryChaplin:
Yeah that's maybe more neat 🤔

AllanOcelot:
@CoryChaplin  since you have a new(er) PR solving the same issue, would you like to close this PR? :)

PR #1528, opened 2021-04-21
Ad container

CWO-Stanley:
Adding an adsense container into the state browser
The container can be disabled by setting showAd: false,
The setting is set default to true.

afbeelding


prawnsalad:
There will be no advertising related things within kiwi itself. If you need to have adverts in your kiwi instance then you will need to modify your version or add some sort of plugin.

PR #1529, opened 2021-04-22
Hide topic from channel list on narrow screens

CoryChaplin:
### Before 
before

### After
Kiwi_IRC_-_The_web_IRC_client


AllanOcelot:
@prawnsalad this looks good to me, any thoughts?

PR #1530, opened 2021-05-02
Improve focus behaviour when moving in history and changing buffers

CoryChaplin:
When moving up or down in history, the focus was at the start of the line. This puts it at the end.
It also adds a default focus on the input line so that users see that they can type right away.


ItsOnlyBinary:
This should no longer be needed with https://github.com/kiwiirc/kiwiirc/pull/1591 being merged

PR #1566, opened 2021-07-15
Create empty object for emojis from config.json && do not hard code emoji file extension

### LINKED ISSUE #1565: Can not overwrite emojis and emoji names are case sensitive
**Describe the bug**
Somewhat between a bug and a feature request. I am trying to use the Kiwi IRC emoj picker with [OpenMoji](https://openmoji.org/).
The png version on a local directory generally works, but there is a capitalization issue of the file names (joypixels uses "f" while Openmoji uses "F").

So I tried overwriting the emoji list via the
```
	"emojis": {
    	    "-__-": "1F611",
	    "=$": "1F633",
    	    ":-D": "1F605"
	},
```
Config parameter, but this just appends and not replaces the emojis in the emoji picker.

See also: https://github.com/kiwiirc/kiwiirc/issues/1083 

**To Reproduce**
1. Use OpenMoji as emoji location
2. use emoji config parameter to respecify already known emojis
3. See append only and non working old emojis due to file name capitalisation problem

**Expected behavior**
It would be nice if KiwiIRC would just ignore file name capitalization for the emoji .pngs and also allow overwriting the default emojis in the emoji picker.

ItsOnlyBinary:
Closes #1565


westor7:
I've tested this PR and is working great without any problems, i think it's good and ready for merge.

westor7 (approved):
I've tested this PR and is working great without any problems, i think it's good and ready for merge.

PR #1568, opened 2021-07-18
Updated translation

Felipefpl:
Updated translation to brazilian portuguese


Shillos:
@Felipefpl kiwiirc project uses crowdin.com for the translations and is not directly with the translations in files.

Felipefpl:
thanks for the tip, i submitted on crowdin. :)

Shillos:
Just some tips: It looks from the file of the translation you must add the {{ }} as is so hopefully you will accordingly like for example {{ nick }} that stays as is its never translated anything that is command-based or within those brackets should not be changed. Another thing the first letter should be a capital letter as I saw on your document you literally changed it to lower case, that should not be done. Just translate the bit that is not translated otherwise the rest is ok.

PR #1577, opened 2021-07-30
Wait for nick change event in SelfUser to correctly show errors

### LINKED ISSUE #1401: issue in the block to change nick
when you choose to change your nickname if you do not fill in the empty space and type accept, that window is locked obstructing you to press any button again

ItsOnlyBinary:
closes #1401
closes #1575


ItsOnlyBinary:
- Add a loader to indicate something is happening
- Add a timeout in case of unexpected event
- Double check if the remove form was there for styling

Issue #1457, opened 2020-12-28
 Option to disable autocompletion of commands


MilitumChristi:
When you type "/" on the keyboard when entering text in the chat, it offers the available commands like: / action / away / back / clear / ctcp / cycle / dice ...

 It would be interesting to be able to have the possibility of deactivating this function, but without deactivating autocompletion for the nicknames.

 There is indeed "showAutoComplete": false, but it blocks all autocompletes.

PR #1583, opened 2021-08-24
Added ShowAutoCompleteCommand option in settings

AnaBatranovic:
As per #1457, added the option to disable the autocomplete function specifically for commands. If the user decides to the disable this setting, the autocomplete popup is automatically closed if the user types /.

### DIFF
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -82,10 +82,6 @@
                                 </select>
                             </label>
                             <label class="u-checkbox-wrapper">
-                                <span>{{ $t('settings_autocomplete_commands') }} </span>
-                                <input v-model="settingAutoCompleteCommands" type="checkbox">
-                            </label>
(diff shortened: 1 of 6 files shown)


mestaritonttu:
Looks like this is now superseded by #1590

ItsOnlyBinary:
@mestaritonttu is correct, I was making other changes in autoComplete and to prevent merge conflict I also implemented this as this pull-request as it was not yet ready for merging.

For future reference this is what I was hoping the pull-request would look like:
https://github.com/kiwiirc/kiwiirc/pull/1590/commits/4b2c5061b6736ca933d545a886eedfaeae5bdda5

After some discussion with people in IRC it was requested that more auto complete tokens where able to be disabled so I instead changed the config option to `showAutocompleteTokens` and implemented here:
https://github.com/kiwiirc/kiwiirc/pull/1590/commits/2ec7390388fc7dd7a9e5a618a327c0cdbc1857a3

Issue #248, opened 2018-03-01
Extra mIRC Colours support


zoltyvigo:
On February 28, the new mIRC 7.52 has released. Its supports 99 colors.

The new colors are the following:

![image](https://user-images.githubusercontent.com/6929524/36842320-a283a26c-1d4b-11e8-9cbf-83d053043af4.png)

AllanOcelot:
@prawnsalad  do you want me to add these into the user colour selection?

prawnsalad:
Sure thing. Though it might be better to select the 52-63 row first, with an option to show more colours.

AllanOcelot:
@prawnsalad agree with that, I'll add it to my list :)

AllanOcelot:
We have replaced the text styling colours to those suggested by @zoltyvigo  in the PR #246

AllanOcelot:
Addressed in #969

zoltyvigo:
New color dialog mIRC 7.62

![image](https://user-images.githubusercontent.com/6929524/90108140-21a4fa80-dd4a-11ea-9de9-43d36f16ca9f.png)

PR #1597, opened 2021-09-30
Make input tools an expanding drawer

ItsOnlyBinary:
image

image


Shillos:
That would look neat, cos on mobile people tell me also they get rather confused with so many buttons on the input, and looks crowded. This looks great and so to have the option for either making it vertical or as compact would help a lot.

Shillos:
Would also be nice to have a more modernized color picker with the latest mIRC colors like in #248

ItsOnlyBinary:
I will do the colour picker at some point, but unlikely before the next release.

I'm going to spend a little time making the drawer a little smarter, as currently it will be shown even if there is only one button which is kinda pointless

ItsOnlyBinary:
combine commits and rebase

Issue #1385, opened 2020-07-06
/ban alias should use the smarter ban masks


prawnsalad:
`/ban` alias passes its arguments directly to the MODE command. It should be smarter and use the same logic as the userbox ban button.

ItsOnlyBinary:
For this issue [createBanMask()](https://github.com/kiwiirc/kiwiirc/blob/master/src/components/UserBox.vue#L340) needs moving to [UserState.js](https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/state/UserState.js), refactored to work and renamed to `getBanMask()`

In [aliases](https://github.com/kiwiirc/kiwiirc/blob/master/src/res/configTemplates.js#L140) ban and unban commands need removing and recreating in [InputHandler.js](https://github.com/kiwiirc/kiwiirc/blob/master/src/libs/InputHandler.js) to use the `getBanMask()` function

**for InputHandler**

The input line should be split by space and first part [0] checked if its a channel name `network.isChannelName(input[0])`, if not the current channel used `state.getActiveBuffer()` `buffer.isChannel()` `buffer.name`

If the nick contains `@` it should be used as the ban mask, if not try to get a user by that name `state.getActiveNetwork()` and `state.getUser(network.id, nick)`

It should require a translation for `"User '{{nick}}' not found!"` for when `/ban non_existant_nick` is used

**example commands that should all work**
`/ban #kiwiirc prawnsalad`
`/ban prawnsalad`
`/ban prawnsalad!*@*`

PR #1601, opened 2021-10-07
Adding Smarter Ban Masks

AnaBatranovic:
As per issue #1385 and the comment left by ItsOnlyBinary, i've refactored the createBanMask() function from Userbox.vue to UserState.js. I've removed the aliases in configTemplate.js, however I'm not sure how to refactor them into InputHandler.js. 

I will update this during the week. Any help would be appreciated. Thanks!

### DIFF
--- a/src/components/AppSettings.vue
+++ b/src/components/AppSettings.vue
@@ -82,10 +82,6 @@
                                 </select>
                             </label>
                             <label class="u-checkbox-wrapper">
-                                <span>{{ $t('settings_autocomplete_commands') }} </span>
-                                <input v-model="settingAutoCompleteCommands" type="checkbox">
-                            </label>
-                            <label class="u-checkbox-wrapper">
                                 <span>{{ $t('settings_timestamps') }} </span>
                                 <input v-model="settingBufferShowTimestamps" type="checkbox">
                             </label>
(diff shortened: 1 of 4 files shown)


ItsOnlyBinary:
You are getting very close to a working pull-request looking at the correct commit.

few tips to help you along:

**UserState,js**
`this.user.account` should be `this.account` as `this` is the user. This mistake is made throughout.

**InputHandler.js**
`if (line === '') {`  trim the whitespace before checking if the line is empty (maybe its all spaces) [line.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim)

`lineParts[0].charAt(0) !== '@'` the format for masks is normally `nick!user@host` so maybe use [String.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

Its also come to mind that there are also extended bans generally of the format $a:account so maybe for that check use:
```
if (lineParts[0].includes('@') || lineParts[0].includes('$')) {
 toBan = lineParts.shift();
} else {
  let nick = lineParts.shift();
  // get user this.state.getUser(network.id, nick);
  // check user exists
  // get banMask
}
```

PR #1673, opened 2022-05-10
Always show hidden messages that affect the user

### LINKED ISSUE #1658: show_joinparts when false should still show the current users join/part/kick
Especially kick as currently when hiding joinparts it also hides the kick and reason

ItsOnlyBinary:
closes #1658


ItsOnlyBinary:
Upon revising this also look into kick visibility, maybe it should use error styling?

PR #1691, opened 2022-08-22
Feat secret channels

CoryChaplin:


ItsOnlyBinary:
Looks good, can you please change the string to "Hide this channel from lists"

feels a little strange it having channel in there twice and its not just channel list is also whois channel list

CoryChaplin:
@ItsOnlyBinary Done ✅

PR #1894, opened 2025-01-06
Use context throughout InputHandler.js

### LINKED ISSUE #1893: "Run commands when connected" commands get sent to the wrong server.
**Describe the bug**
"Run commands when connected" commands get sent to the server window that is open instead of the server where I clicked connect.

**To Reproduce**
1. Open/create multiple IRC connections.
2. Configure "Run commands when connected" commands on at least one.
3. Close the browser window and reopen.
4. Click "connect" on a network connection with "Run commands when connected" commands other than the active one.
5. Notice that the commands were sent to the server that was active and already connected rather than the new connection.

**Expected behavior**
I expected the "Run commands when connected" commands to get sent to the server I clicked connect on.

 - OS: All, but I have tested Windows 11, Mac OS, and Android. 
 - Browser Edge, Chrome, and others.
 - Version Latest available

**Additional context**
If I click to display the message window for the server I am connecting to before clicking connect the commands are sent to the correct server.

ItsOnlyBinary:
This fixes auto commands being sent to active network.

closes #1893

### DIFF
--- a/src/libs/InputHandler.js
+++ b/src/libs/InputHandler.js
@@ -107,13 +107,13 @@ export default class InputHandler {
         };
 
         // Plugins may tap into this event to handle a command themselves
-        this.state.$emit('input.command.' + command, eventObj, command, params);
+        this.state.$emit('input.command.' + command, eventObj, command, params, context);
         if (eventObj.handled) {
             return;
         }
 
         if (inputCommands[command.toLowerCase()]) {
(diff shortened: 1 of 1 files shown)


abcbarryn (approved):
This change seems to fix the issue: "Run commands when connected" commands get sent to the wrong server. #1893
I tested it and it seems to address the issue and not break anything else.

PR #1947, opened 2026-03-06
fix: surface actual error message on socket close instead of err_unknown

CoryChaplin:


ItsOnlyBinary:
This is a hit and miss unfortunately due to the way `network.state_error` works. It requires the error to be a string and then uses [Misc.networkErrorMessage(err)](https://github.com/kiwiirc/kiwiirc/blob/0416529fda6ddf8ab5559796ec69b49e07bd798f/src/helpers/Misc.js#L99) to convert that string into a translated error.

CoryChaplin:
@ItsOnlyBinary OK, this must be better.

PR #1950, opened 2026-03-18
draft CHATHISTORY TARGETS support

### LINKED ISSUE #1923: Private Message chatHistory targets Not Loading on Reconnect if someone sent a private message while offline
## **Problem Description / Bug**
KiwiIRC currently supports chathistory and buffers.auto_request_history (including the "all" setting). However, there appears to be a major flaw regarding private messages (PMs) received on IRCv3 compatible networks such as Ergo:

If a user receives a private message while they are offline or do not have an active private message buffer open for that specific sender, the message history for that PM is not retrieved or displayed upon subsequent login/reconnect.

This means that users are unable to review past private messages from senders for whom they did not have an open PM buffer at the time of message reception. This effectively makes it impossible to know who sent a PM or what its content was if the buffer wasn't already open while the user was disconnected. Or once reconnecting to a brand new kiwiirc session..

## Steps to Reproduce (Scenario)

1.   Connect to an ERGO IRC network using KiwiIRC.
2.   Ensure buffers.auto_request_history is set to "all" in the configuration.
3.   Have another user send you a private message while you   Offline (disconnected from KiwiIRC).
4.  Log back into KiwiIRC as the second account.
5.  No automatic PM window opens.  

**Expected Result**: The private message history for that sender should be retrieved and displayed, including messages sent while offline or without an open buffer.

Actual Result: The private message is not loaded, and there is no indication that a message was received from that sender unless a new message is sent, or the history is manually requested ex by opening a manual buffer/message. But this is highly unrealistic as you don't know someone sent a message

**Attempts and experiments**
I have tried `configuring buffers.auto_request_history` to "all" and also other settings, but this did not resolve the issue for private messages.

### Impact

This is a significant functional blocker for daily IRC usage, especially on networks leveraging chathistory capabilities. It severely impacts user experience and can lead to missed communications. While alternative clients like Gamja currently handle private message history retrieval correctly in this scenario, KiwiIRC's overall feature set and user experience are highly preferred.

Resolving this issue would greatly enhance KiwiIRC's robustness and bridge a crucial gap, contributing positively to broader IRC adoption.

- Linux Mint/ Debian
- Both FireFox, LibreWolf and Chromium

**Smartphone (please complete the following information):**
- not applicable

Thank you for your attention to this matter.

slingamn:
Fixes #1923. Vibecoded with Claude Code; I did some cursory testing in a browser.

### DIFF
--- a/src/libs/ChathistoryMiddleware.js
+++ b/src/libs/ChathistoryMiddleware.js
@@ -18,14 +18,30 @@ export default function chathistoryMiddleware() {
     };
 
     function theMiddleware(command, event, client, next) {
-        if (command.toLowerCase() === 'batch end chathistory' && client.chathistory) {
-            let target = event.params[0];
-            client.chathistory.batchCallbacks.resolve(target, event);
+        if (command === 'batch end' && client.chathistory) {
+            if (event.type === 'chathistory') {
+                let target = event.params[0];
+                client.chathistory.batchCallbacks.resolve(target, event);
+            } else if (event.type === 'draft/chathistory-targets') {
+                if (client.chathistory.pendingTargets) {
+                    let resolve = client.chathistory.pendingTargets;
+                    client.chathistory.pendingTargets = null;
+                    resolve(event);
+                }
+            }
         }
 
         // This is sent as 'unknown command', its been structured this way so hopefully it
         // will still work if a fail handler is ever created in irc-fw
         if (event?.command?.toLowerCase() === 'fail' && event.params[0].toLowerCase() === 'chathistory') {
+            // If this FAIL is for a TARGETS request, resolve the pending targets promise
+            if (client.chathistory.pendingTargets && event.params[2]?.toUpperCase() === 'TARGETS') {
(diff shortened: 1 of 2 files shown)


src/libs/ChathistoryMiddleware.js
```diff
@@ -18,14 +18,30 @@ export default function chathistoryMiddleware() {
...
 
     function theMiddleware(command, event, client, next) {
-        if (command.toLowerCase() === 'batch end chathistory' && client.chathistory) {
-            let target = event.params[0];
-            client.chathistory.batchCallbacks.resolve(target, event);
+        if (command === 'batch end' && client.chathistory) {
+            if (event.type === 'chathistory') {
+                let target = event.params[0];
+                client.chathistory.batchCallbacks.resolve(target, event);
+            } else if (event.type === 'draft/chathistory-targets') {
```
slingamn, line 25: I think it would actually be appropriate to future-proof this immediately, and accept `chathistory-targets` as well

PR #1954, opened 2026-04-12
Improve two character language detection

### LINKED ISSUE #1953: Preferred language misdetection
The code responsible for detecting the user's preferred language from browser headers has a bug:

https://github.com/kiwiirc/kiwiirc/blob/3836a252fd95b09abe6823f05895b97cb525fd08/src/main.js#L470-L474

My browser's preferred languages are English and French, in that order, which results in the header `Accept-Language: en,fr;q=0.9`. The code above first tries `en-en`, which doesn't exist, then `fr-fr`, which does, and so the interface is displayed in French, against my preference.

ItsOnlyBinary:
closes #1953

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -470,7 +470,8 @@ function initLocales() {
             // if this is a language such as 'fr', add a following one of 'fr-fr' to cover
             // both cases
             if (lang.length === 2) {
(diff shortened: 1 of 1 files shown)


ncfavier:
Thanks, looks reasonable!

PR #1962, opened 2026-05-10
Add feature to strip colors on incoming messages

Kajvdh:
Some of our users have colorblindness and were requesting a client-side setting to be able to strip colors on incoming messages.

I figured to create a PR for this in case others would like this as well


ItsOnlyBinary:
can you please revert all the locale changes except for app.dev.po
translations are handled by crowdin and im not sure what will happen if i merge this with changes to the other locales

guest20:
It looks like the platform is supposed to magically discover the translations ... 

https://crowdin.com/blog/continuous-localization#continuous-localization-workflow

